Issue #289, opened 2026-01-05
`bakery update files --template-pattern` flag does not work


ianpittwood:
The `--template-pattern` filter flag does not correctly isolate which templates it is rendering.

PR #298, opened 2026-01-07
Fixes for `imagetools create` cross-registry authentication failures

### LINKED ISSUE #294: Stop `bakery clean` commands from failing fast on request errors
Any non-successful response from the Github API for managing our image artifacts currently results in an instant exit and failure for cleanup. We should modify the behavior to fail after all queued calls are made and print the error messages for failed items as we go.

ianpittwood:
Fixes #294, #289

- Overhaul the image merge process to bypass the cross-registry authentication issues we encountered in the above issues.
- Add service for local in-memory registry for temporary image storage.
- Add custom image inspection utility function to support index artifacts.
- Surface log level in settings.

### DIFF
--- a/posit-bakery/posit_bakery/image/image_metadata.py
+++ b/posit-bakery/posit_bakery/image/image_metadata.py
@@ -21,6 +21,8 @@ class ImageToolsInspectionPlatformMetadata(BaseModel):
         if self.architecture:
             s += self.architecture
 
+        s = s.rstrip("/")
+
         return s
 
 


posit-bakery/posit_bakery/image/image_metadata.py
```diff
@@ -5,6 +5,44 @@
...
+    architecture: Annotated[str | None, Field(description="The architecture of the built image.", default=None)]
+    os: Annotated[str | None, Field(description="The operating system of the built image.", default=None)]
+
+    def __str__(self):
+        """Returns a string representation of the platform."""
+        s = ""
+        if self.os:
+            s += self.os + "/"
+        if self.architecture:
+            s += self.architecture
```
bschwedler: If `architecture is None`, this could conceivably end with a trailing slash.

bschwedler (approved):
The process, although convoluted, makes sense to me.


I am curious how much extra time pulling/pushing all the images locally adds.

Issue #306, opened 2026-01-14
Make template rendering compatible with `ImageMatrix` models


ianpittwood:
Depends on #152 

We need to bridge our logic for rendering/managing templates into creating buildable definitions that are easily usable by the `ImageMatrix` definition. The goal remains to create something that `bakery` can easily call programatically for builds/tools while also allowing for a directly buildable `Containerfile` definition that can be utilized without `bakery`.

ianpittwood:
My initial idea is to have templates that can render to a single `Containerfile` that utilizes non-defaulted build args in order to build out the correct combinations of components. This should allow a customer to build out an image with custom versions outside of `bakery` while also giving us an easy way to programatically perform builds/tools calls within `bakery.

PR #308, opened 2026-01-14
Add build args capabilities into dependency macros

ianpittwood:
Related to #306 

- Adds macros for defining and referencing build args for dependencies.

### DIFF
--- a/posit-bakery/posit_bakery/config/dependencies/__init__.py
+++ b/posit-bakery/posit_bakery/config/dependencies/__init__.py
@@ -8,7 +8,6 @@ from .python import PythonDependencyConstraint, PythonDependencyVersions
 from .quarto import QuartoDependencyConstraint, QuartoDependencyVersions
 from .r import RDependencyConstraint, RDependencyVersions
 
-
 DependencyConstraintField = Annotated[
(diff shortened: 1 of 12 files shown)


posit-bakery/posit_bakery/config/templating/macros/python.j2
```diff
@@ -1,5 +1,15 @@
...
 
+{# Declare a generic build argument for a Python version #}
+{% macro declare_build_arg(name = "PYTHON_VERSION", default = "") -%}
+ARG {{ name }}{% if default %}={{ default }}{% endif %}
+{%- endmacro %}
+
+{# Reference the build argument for a Python version #}
+{% macro build_arg(name = "PYTHON_VERSION") -%}
+${{ name }}
+{%- endmacro %}
```
bschwedler: Is this for convenience in case we change the name?

Given this function has the `name` parameter, could itconceivably be invoked by something like?
```
{{ python.build_arg(name="JAVA_VERSION") }}
```

Would it make sense to have a const string that we reference in these macros?

ianpittwood: I changed to using a constant. I get nervous about all the contexts and stuff with these values possible overlapping, but I think we'll be fine. It's passing tests with the change.

PR #314, opened 2026-01-20
Update retrieval of Package Manager dev versions

bschwedler:
Package Manager has moved to using the same version string format that
connect uses.

Example: `2026.01.0-dev+167-gd27bbec1d7`


ianpittwood (approved):
LGTM

PR #336, opened 2026-01-27
Refactor image version and matrix

bschwedler:
This change attempts to consolidate logic between the image version and image matrix.

I used claude interactively, developing plans, reviewing, iterating, and making multiple changes that resulted in a single commit.

### DIFF
--- a/posit-bakery/posit_bakery/config/image/version_matrix_base.py
+++ b/posit-bakery/posit_bakery/config/image/version_matrix_base.py
@@ -1,0 +1,292 @@
+"""Shared methods for ImageVersion and ImageMatrix."""
+
+import abc
+import logging
+import re
+from copy import deepcopy
+from pathlib import Path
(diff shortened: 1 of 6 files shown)


posit-bakery/posit_bakery/config/image/version_matrix_base.py
```diff
@@ -0,0 +1,292 @@
...
+    @abc.abstractmethod
+    def _get_image_template_values(self) -> dict[str, Any]:
+        """Returns image-specific template values to add to the Image dict.
+
+        :return: Dictionary with values like Version, IsDevelopmentVersion, etc.
+        """
+        ...
+
+    @property
+    def path(self) -> Path | None:
```
ianpittwood: I believe this is already provided by `BakeryPathMixin`? Perhaps there's more we can share with additional classes using that mixin and a rename.

PR #357, opened 2026-02-11
Add related directories/project to Claude settings

bschwedler:
The images project is spread across multiple repositories.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,8 +1,15 @@
 {
-  "lsp": {
-    "pyright": true
+  "enabledPlugins": {
+    "pyright-lsp@claude-plugins-official": true
   },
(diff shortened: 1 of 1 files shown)


.claude/settings.json
```diff
@@ -1,8 +1,15 @@
...
+  "enabledPlugins": {
+    "pyright-lsp@claude-plugins-official": true
   },
   "permissions": {
+    "additionalDirectories": [
+      "../images",
+      "../images-examples",
+      "../images-connect",
+      "../images-package-manager",
+      "../images-workbench"
```
ianpittwood: The theory here is that you would have the images all cloned to the same base directory, am I understanding right?

ianpittwood (approved):

PR #373, opened 2026-03-10
Add --pull flag to bakery build to always pull latest base images

ianpittwood:
Fixes CI failures where cached UV images don't have metadata for newly released Python versions. The --pull flag is passed through both bake and sequential build strategies. CI workflows always use --pull.


bschwedler (approved):
This looks good.

Later we can decide to make this an option/input to the workflow.

PR #432, opened 2026-04-08
Add Slack build notification action

bschwedler:
## Summary

Adds a reusable composite action at `.github/actions/slack-build-notify/` that sends Slack notifications on build state transitions on `main`:

- **Failure**: notifies when a previously-passing workflow starts failing, with failed job/step details and links
- **Recovery**: notifies when a previously-failing workflow recovers, with consecutive failure count

Designed for use in the CI meta-job across product image repos (`images-connect`, `images-package-manager`, `images-workbench`).

### Design decisions

- Uses `actions/github-script` for pre-authenticated GitHub API access
- Compares the previous run's CI job conclusion (not workflow conclusion) to avoid false notifications from clean-job flaps
- Graceful failure: try/catch + `continue-on-error` in callers so notification issues never block CI
- Restricted to `posit-dev/` repos via `github.repository_owner` guard
- Slack Block Kit messages with `text` fallback for notification previews

### Required setup

Each product repo needs a `SLACK_WEBHOOK_URL` repository secret pointing at the `#platform-builds-images` channel webhook.

## Test plan

- [x] Add `SLACK_WEBHOOK_URL` secret to a test repo
- [ ] Trigger a failure on `main` and verify Slack message appears
- [ ] Trigger a recovery and verify consecutive failure count is correct
- [ ] Verify no notification fires when state doesn't change
- [ ] Verify no notification fires on PR branches

### DIFF
--- a/.github/actions/slack-build-notify/action.yml
+++ b/.github/actions/slack-build-notify/action.yml
@@ -1,0 +1,197 @@
+name: Slack Build Notification
+description: |
+  Sends Slack notifications on build state transitions on main.
+  Notifies when a previously-passing workflow starts failing,
+  and when a previously-failing workflow recovers.
+
+inputs:
+  result:
+    description: Current build result (success or failure)
+    required: true
+  slack-webhook-url:
+    description: Slack incoming webhook URL
+    required: true
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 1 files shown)


ianpittwood (approved):
LGTM but some of the Copilot feedback may be actionable

PR #438, opened 2026-04-10
Scope clean workflows to match build version filters

bschwedler:
## Summary

- Add `--dev-versions` and `--matrix-versions` flags to `bakery clean cache-registry` and `bakery clean temp-registry`
- Pass them through as inputs on `clean.yml`
- Remove stale FIXME in `bakery-build.yml` (QEMU workflow only used for CI self-tests)

Previously, clean jobs only loaded production image targets (the default), so dev and matrix version caches accumulated forever. Sibling PRs in each image repo pass the matching flags.

## Test plan

- [x] 886 config tests pass
- [ ] `bakery clean cache-registry --help` shows `--dev-versions` and `--matrix-versions`
- [ ] `bakery clean temp-registry --help` shows `--dev-versions` and `--matrix-versions`

### DIFF
--- a/.github/workflows/bakery-build.yml
+++ b/.github/workflows/bakery-build.yml
@@ -160,7 +160,6 @@ jobs:
       - name: Build
         env:
           GIT_SHA: ${{ github.sha }}
-        # FIXME: Currently pushes to ghcr.io for caching. Needs to be conditional
         run: |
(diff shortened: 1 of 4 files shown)


ianpittwood (approved):
I think this should fix the issue.

PR #469, opened 2026-04-16
Propagate template file mode to rendered output

bschwedler:
## Summary

- Bakery's render path used `open(path, "w") + f.write(rendered)`, which created rendered files with the default mode (0o644 after umask). The exec bit on a template was never reflected in its rendered file — shell scripts rendered from `.sh.jinja2` templates had to be chmod +x'd manually each render.
- Swap to `copy2(tpl, out); out.write_text(rendered)`. `copy2` preserves source mode (and timestamps, which `write_text` then overwrites), mirroring the idiom already used on the verbatim non-template-copy branch (`version.py:410`). No new imports.
- Covers both the `ImageVersion` and `ImageMatrix` render paths (4 write sites total).

Backwards-compatible: a 0o644 template continues to render to a 0o644 output; only chmod +x templates change behavior.

This unblocks sibling-repo cleanup where the pre-commit `check-shebang-scripts-are-executable` hook flags rendered `.sh` files with shebangs that aren't executable. After this lands, marking a `.sh.jinja2` template +x is enough — re-render propagates.

## Test plan

- [x] New unit test `test_render_files_preserves_template_file_mode` asserts that a 0o755 template renders to a 0o755 file and a 0o644 template stays 0o644.
- [x] Existing render_files tests in `test_version.py` and `test_matrix.py` still pass.
- [x] Smoke test: in a sibling repo (e.g. images-package-manager), chmod +x `package-manager/template/scripts/*.jinja2`, run `bakery update files`, verify rendered `.sh` files come out executable.

### DIFF
--- a/posit-bakery/posit_bakery/config/image/matrix.py
+++ b/posit-bakery/posit_bakery/config/image/matrix.py
@@ -522,9 +522,9 @@ class ImageMatrix(BakeryPathMixin, BakeryYAMLModel):
                                 )
                             )
                             continue
-                        with open(containerfile, "w") as f:
-                            log.debug(f"[bright_black]Rendering [bold]{containerfile}")
-                            f.write(rendered)
+                        log.debug(f"[bright_black]Rendering [bold]{containerfile}")
+                        copy2(tpl_full_path, containerfile)
+                        containerfile.write_text(rendered)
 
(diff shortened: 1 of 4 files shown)


ianpittwood (approved):
LGTM

PR #470, opened 2026-04-16
Enforce pre-commit in CI

bschwedler:
## Summary

- Adds a `lint` job to `ci.yml` that runs pre-commit on every push/PR/merge_group, gated by the meta `ci` job's `needs:` list so hook failures block merges
- Swaps `beautysh` → `shfmt` in `.pre-commit-config.yaml` (beautysh is unmaintained and broken on Python 3.12)
- Removes a commented-out mypy block and the beautysh setuptools workaround
- Inline lint job: SHA-pinned to `actions/checkout@v6.0.2`, `actions/setup-python@v5.6.0`, `pre-commit/action@v3.0.1`; `permissions: contents: read`; `SKIP: no-commit-to-branch` env

No reusable workflow — the job is small enough that inlining avoids cross-repo merge-order dependencies.

## Test plan

- [x] `pre-commit run --all-files` passes locally on this branch
- [x] `lint` job in CI passes on this PR
- [x] Zizmor job passes (SHA pins satisfy the hash-pin policy)


.github/workflows/ci.yml
```diff
@@ -38,6 +39,23 @@ jobs:
...
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
+        with:
+          python-version: "3.x"
```
ianpittwood: Can we point to the pyproject.toml instead?

ianpittwood (approved):

PR #467, opened 2026-04-16
Deduplicate dev version filter logic

bschwedler:
## Summary

Extracts `matches_dev_filter()` method on `ImageVersion` to deduplicate the dev version filtering logic that was duplicated between `generate_image_targets` (config.py) and `ci matrix` (ci.py).

Follow-up to #457.

### Changes
- Add `ImageVersion.matches_dev_filter(dev_versions, dev_stream)` returning `(bool, reason)`
- `generate_image_targets` uses the reason for warning/debug messages
- `ci matrix` uses the bool, discards the reason
- Fix `ci merge --dev-stream` help panel: "Build Configuration & Outputs" → "Filters"
- 11 parametrized tests for `matches_dev_filter` covering all combinations
- Test for `--dev-stream` with `dev_versions=INCLUDE` (mixed dev + release)

## Test plan

- [x] 918 config tests pass (11 new)

Issue #302, opened 2026-01-07
Add `workflow_dispatch` support to image ci workflows


bschwedler:
We want the ability to trigger the ci workflow for each image via the [workflow_dispatch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?versionId=free-pro-team%40latest&productId=actions&restPage=how-tos%2Cwrite-workflows%2Cchoose-when-workflows-run%2Ctrigger-a-workflow#workflow_dispatch) trigger.

This will allow us to use a GitHub App and trigger with [gh workflow run](https://cli.github.com/manual/gh_workflow_run) with the `--repo` flag.

bschwedler:
### Prerequisites

- [x] Merge release workflow PRs (#393, posit-dev/images-connect#41, posit-dev/images-workbench#49, posit-dev/images-package-manager#55)
- [x] Set up a GitHub App with `actions:write` on the `posit-dev/images-*` and `posit-dev/helm` repos
- [x] Switch preview images from ECR to GHCR (posit-dev/images-workbench#54, posit-dev/images-package-manager#57)

### Per-product work

#### Connect

**Image repo** (`posit-dev/images-connect`):
- Add `workflow_dispatch` trigger to `development.yml` and `production.yml`
- After a successful production build, dispatch a workflow in `posit-dev/helm` with the product name, version, and image tag

**Product repo** (`posit-dev/connect`):
- Add a step to [`tools/release/publish_release.py`](https://github.com/posit-dev/connect/blob/main/tools/release/publish_release.py) calling `run_github_actions_workflow()` to dispatch `release.yml` in `posit-dev/images-connect`. The function already exists in [`release_common.py`](https://github.com/posit-dev/connect/blob/main/tools/release/release_common.py) and is used for 3 other dispatch calls (docs, sbom, escrow).

#### Workbench

**Image repo** (`posit-dev/images-workbench`):
- Add `workflow_dispatch` trigger to `development.yml`, `production.yml`, and `session.yml`
- After a successful production build, dispatch a workflow in `posit-dev/helm` with the product name, version, and image tag

**Product repo** (`rstudio/rstudio-pro`):
[...]

PR #480, opened 2026-04-22
Add image-version and dev-stream inputs to build workflows

bschwedler:
## Summary

Dispatches from product repos need to narrow the build matrix to a specific image version and release stream. Main's `bakery build` and `bakery run dgoss` already accept `--image-version` and `--dev-stream` (the CLI half landed in #457), but the reusable workflows don't expose those inputs yet — so the three in-flight `dev-dispatch-inputs` PRs in images-connect/images-workbench/images-package-manager have to pin to the orphaned `release-stream-filter` branch.

This PR closes that gap by adding two optional inputs (`image-version`, `dev-stream`) to `bakery-build-native.yml` and `bakery-build.yml`, and plumbing them through to the matrix and build steps. It also folds in #467 so the `ci matrix` loop uses the shared `matches_dev_filter()` method instead of inline dev/stream checks.

## Changes

- `bakery-build-native.yml` and `bakery-build.yml`: add `image-version` and `dev-stream` as optional `workflow_call` inputs (default empty). Append them to `bakery ci matrix` calls only when set, using bash arrays so existing callers keep working unchanged.
- `posit_bakery/cli/ci.py`: add `--image-version` option to `bakery ci matrix`. Without this, passing `--image-version` to matrix generation would fail on main (only `bakery build` and `bakery run dgoss` had it). Exits non-zero when the filter matches nothing so CI doesn't silently skip all build jobs.
- Folded from #467: `ImageVersion.matches_dev_filter()` extracted to deduplicate the dev/stream check between `generate_image_targets` and `ci matrix`; Pydantic `model_fields` deprecation fix on `dependency.py`.
- Feature scenarios for matching, non-matching, and combined `--image-version` filters.

Scope intentionally excludes `dev-channel` / `--value` overrides from the `release-stream-filter` branch — the bakery CLI has no `--dev-channel` on main and the dispatch PRs listed in #302 don't use it.

## Unblocks

- https://github.com/posit-dev/images-connect/pull/66
- https://github.com/posit-dev/images-workbench/pull/74
- https://github.com/posit-dev/images-package-manager/pull/69

After this merges, those PRs can repoint their `development.yml` from `bakery-build-native.yml@release-stream-filter` to `@main`.

## Supersedes

- #467 (folded into this PR)

## Test plan

- [x] Existing `bakery ci matrix` tests pass (unchanged).
- [x] New `--image-version` filter produces a matrix containing only the named version.
- [x] Non-matching `--image-version` exits non-zero.
- [x] 924 bakery tests pass locally.

### DIFF
--- a/posit-bakery/posit_bakery/config/image/version.py
+++ b/posit-bakery/posit_bakery/config/image/version.py
@@ -11,10 +11,11 @@ from pydantic import Field, field_validator, model_validator
 from pydantic_core.core_schema import ValidationInfo
 
 from posit_bakery.config.dependencies import DependencyVersionsField
+from posit_bakery.config.image.posit_product.const import ReleaseStreamEnum
 from posit_bakery.config.registry import BaseRegistry
 from posit_bakery.config.registry import Registry
 from posit_bakery.config.shared import BakeryPathMixin, BakeryYAMLModel
-from posit_bakery.const import JINJA2_TEMPLATE_EXTENSIONS
+from posit_bakery.const import DevVersionInclusionEnum, JINJA2_TEMPLATE_EXTENSIONS
 from .build_os import DEFAULT_PLATFORMS, TargetPlatform
 from .variant import ImageVariant
 from .version_os import ImageVersionOS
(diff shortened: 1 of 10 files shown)


posit-bakery/posit_bakery/config/image/version.py
```diff
@@ -119,6 +120,27 @@ class ImageVersion(BakeryPathMixin, BakeryYAMLModel):
...
+        """
+        if self.isDevelopmentVersion and dev_versions == DevVersionInclusionEnum.EXCLUDE:
+            return False, "excluded by --dev-versions exclude"
+        if not self.isDevelopmentVersion and dev_versions == DevVersionInclusionEnum.ONLY:
+            return False, "not a development version (excluded by --dev-versions only)"
+        if dev_stream is not None and self.isDevelopmentVersion:
+            version_stream = self.metadata.get("release_stream")
+            if version_stream != dev_stream:
+                return False, (f"dev stream '{version_stream}' does not match --dev-stream '{dev_stream.value}'")
+        return True, None
```
ianpittwood, line 142: Claude, this is terrible and you know it.

ianpittwood (approved):
LGTM, tested locally against PPM

PR #486, opened 2026-04-24
Add buildSecrets config for passing secrets to image builds

ianpittwood:
## Summary
- Adds `Image.buildSecrets` — a list of `{id, envVar}` entries — that is translated into `--secret id=<id>,env=<envVar>` options for sequential `docker buildx build` and into `{type: env, id, env}` per-target entries for Docker Bake plans.
- `BuildSecret.id` and `BuildSecret.envVar` are pattern-validated to safe character sets (alphanumerics + `_`/`.`/`-` for ids; POSIX env-var names for envVar) to prevent CLI argument injection like `id=foo,src=/etc/passwd`.
- Resolution lives in a single `ImageTarget.resolved_build_secrets` property; sequential and bake paths call `as_cli_option()` / `as_bake_json()` on the resolved entries.
- Secrets whose `envVar` is unset are skipped with a warning; required-secret enforcement stays with Docker via `RUN --mount=type=secret,id=<id>[,required=true]` in the Containerfile.
- Addresses #483 — enables product repos to inject `GITHUB_TOKEN` into `quarto install tinytex` (and similar) to avoid unauthenticated GitHub API rate limiting.

## Example config
```yaml
images:
  - name: connect-content
    buildSecrets:
      - id: github_token
        envVar: GITHUB_TOKEN
```

Containerfile usage:
```dockerfile
RUN --mount=type=secret,id=github_token,required=false \
    GH_TOKEN="\$(cat /run/secrets/github_token 2>/dev/null || true)" \
    quarto install tinytex
```

## Test plan
- [x] Unit tests cover the BuildSecret model (incl. pattern rejection of injection-flavored values), Image.buildSecrets parsing, and resolution behavior on both build paths.
- [x] Verify sequential build in a product repo with \`GITHUB_TOKEN\` set — secret appears in \`docker buildx build --secret ...\` args.
- [x] Verify bake plan JSON includes per-target \`secret: [{type: env, id: ..., env: ...}]\` entries.
- [x] Verify a missing \`envVar\` logs a warning and does not break the build (when the Containerfile mount uses \`required=false\`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #487, opened 2026-04-24
Mount GitHub token secret for Quarto TinyTeX install

### LINKED ISSUE #483: Mount github token as a secret to builds
# Problem Summary

We currently have flaky build failures that stem from GitHub unauthenticated API request ratelimiting. Whenever we call `quarto install tinytex`, it triggers two separate calls to the GitHub API. These calls are currently unauthenticated as the GitHub Token in Actions is not inherited down into the builder environment and the Quarto CLI looks for `GH_TOKEN` rather than Actions' `GITHUB_TOKEN` anyways. Unauthenticated requests are capped at 60 req/hr while authenticated are capped at 50000 req/hr.

# Proposed Solution

We can inject token into our builds and make it available by doing the following:

1. Add the option `--secret id=github_token,env=GITHUB_TOKEN` to the `docker buildx build` call. Note, this will fail if `GITHUB_TOKEN` is unset in the environment. This can be circumvented by setting `GITHUB_TOKEN=${GITHUB_TOKEN:-}` before the command, but logic in the Dockerfile or any script must account for and ignore empty values.
2. Add `--mount=type=secret,id=github_token,required=false` as an option on any `RUN` step that requires the GitHub Token to be passed in. The secret is mounted to `/run/secrets/github_token` only when this is set for the `RUN` step. Note that setting `required=false` is optional, but would make it so the secret being missing does not immediately fail the build. It can be optional in the case of `quarto install tinytex` or other GitHub API ops.
3. Set the necessary environment variable by setting `GH_TOKEN="$(cat /run/secrets/github_token)"` before commands as needed.

The question becomes how do we want Bakery to handle passing in a secret. There's a couple of viable approaches for implementation in Bakery:

1. Always add `--secret id=github_token,env=GITHUB_TOKEN` whenever `GITHUB_TOKEN` is set regardless of build. This is the easiest, but it is rigid and will often be unnecessary and unused. I don't believe it is a security risk as it is still passed as a secret and the `GITHUB_TOKEN` used by Actions has a short TTL.
2. Add a configuration element to `Image` named `buildSecrets`. This can be a list of maps that set both an `id` and `env`. This allows for flexibility for future secret use cases, but is slightly harder to implement. We would also need to decide if we should always check for `env` being set and possibly fail a build before calling `docker buildx build`. I think it makes most sense to append the `--secret` option conditionally on the set `env` being present, propagate a warning if it is missing, and then defer to `docker buildx build` to fail if the secret was required.

ianpittwood:
## Summary
- Pass an optional `--mount=type=secret,id=github_token,required=false` on the `RUN` step of `quarto.run_install` when `with_tinytex=True`.
- Export `GH_TOKEN` from `/run/secrets/github_token` (when non-empty) inside `install_tinytex_command` so the Quarto TinyTeX install can authenticate to GitHub and avoid rate limits.
- Update `test_install_tinytex_command`, `test_install`, and `test_run_install` expectations accordingly.
- Addresses #483 by adding more helpers in templates.

## Test plan
- [x] `uv run pytest test/config/templating/test_macros.py -k "quarto or Quarto or tinytex"` passes
- [x] Downstream product repos rebuild with `bakery update files` and a real build exercises the secret mount

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
+++ b/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
@@ -23,6 +23,12 @@ ${{ name }}
 {# The --mount option for the optional GitHub token build secret consumed by
    the TinyTeX install step. Mount is declared on the RUN instruction; the
    corresponding shell-side read lives in `install_tinytex_command`.
+
+   By setting GH_TOKEN to a valid GitHub token, we avoid hitting GitHub API
+   rate limits during the TinyTeX install process, which downloads packages
+   from GitHub repositories. The secret is optional to allow for flexibility
+   in build environments, but if not provided, running a high number of
+   concurrent builds that install Quarto may result in flaky failures.
 #}
 {% macro github_token_secret_mount() -%}
 --mount=type=secret,id=github_token,required=false
@@ -35,8 +41,7 @@ ${{ name }}
 :param home_path: An optional path to set the HOME environment variable for the command, which controls where TinyTeX is installed. If not provided, TinyTeX will be installed to the default location under the current user's home directory.
 #}
 {% macro install_tinytex_command(quarto_binary, update_path = False, home_path = None) -%}
-if [ -s /run/secrets/github_token ]; then export GH_TOKEN="$(cat /run/secrets/github_token)"; fi && \
-{% if home_path %}HOME="{{ home_path }}" {% endif %}{{ quarto_binary }} install tinytex --no-prompt --quiet{% if update_path %} --update-path{% endif %}
+GH_TOKEN="$([ -s /run/secrets/github_token ] && cat /run/secrets/github_token)" {% if home_path %}HOME="{{ home_path }}" {% endif %}{{ quarto_binary }} install tinytex --no-prompt --quiet{% if update_path %} --update-path{% endif %}
 {%- endmacro %}
 
 {# Command to download and install a specific version of Quarto, optionally installing TinyTeX as well
--- a/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
+++ b/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
@@ -27,6 +27,7 @@ ${{ name }}
 :param home_path: An optional path to set the HOME environment variable for the command, which controls where TinyTeX is installed. If not provided, TinyTeX will be installed to the default location under the current user's home directory.
 #}
 {% macro install_tinytex_command(quarto_binary, update_path = False, home_path = None) -%}
+if [ -s /run/secrets/github_token ]; then export GH_TOKEN="$(cat /run/secrets/github_token)"; fi && \
 {% if home_path %}HOME="{{ home_path }}" {% endif %}{{ quarto_binary }} install tinytex --no-prompt --quiet{% if update_path %} --update-path{% endif %}
 {%- endmacro %}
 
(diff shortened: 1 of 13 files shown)


posit-bakery/posit_bakery/image/bake/bake.py
```diff
@@ -127,6 +139,20 @@ def from_image_target(
...
+        if image is not None and image.buildSecrets:
+            secrets = []
+            for secret in image.buildSecrets:
+                if os.environ.get(secret.envVar):
+                    secrets.append({"type": "env", "id": secret.id, "env": secret.envVar})
+                else:
+                    log.warning(
+                        f"Build secret '{secret.id}' for image '{image.name}' skipped: environment "
+                        f"variable '{secret.envVar}' is not set."
+                    )
```
bschwedler: This check-and-warn loop is duplicated almost verbatim in `ImageTarget.build_secrets` (`image_target.py`). If either copy gets a bug fix (e.g. the `os.environ.get` truthiness issue), the other silently drifts.

Consider extracting a shared resolver -- either a classmethod on `BuildSecret` or a standalone helper -- so both the bake and sequential paths call the same code:

```python
# On Image or as a standalone helper:
def resolve_build_secrets(self) -> list[BuildSecret]:
    resolved = []
    for secret in self.buildSecrets:
        if secret.envVar in os.environ:
            resolved.append(secret)
        else:
            log.warning(...)
    return resolved
```

Then `BakeTarget.from_image_target` maps to dicts and `ImageTarget.build_secrets` maps to CLI strings -- but the env-var check and warning live in one place.

posit-bakery/posit_bakery/config/templating/macros/quarto.j2
```diff
@@ -20,13 +20,22 @@ ${{ name }}
...
+{%- endmacro %}
+
 {# Command to install TinyTeX using the Quarto CLI
 
 :param quarto_binary: The path to the Quarto binary.
 :param update_path: Adds --update-path to the command to add TinyTeX binaries to the PATH. Defaults to False.
 :param home_path: An optional path to set the HOME environment variable for the command, which controls where TinyTeX is installed. If not provided, TinyTeX will be installed to the default location under the current user's home directory.
 #}
 {% macro install_tinytex_command(quarto_binary, update_path = False, home_path = None) -%}
+if [ -s /run/secrets/github_token ]; then export GH_TOKEN="$(cat /run/secrets/github_token)"; fi && \
```
bschwedler: ``export`` puts ``GH_TOKEN`` into the environment for every subsequent command in this RUN instruction. Even though ``quarto install tinytex`` is currently the only consumer, a future edit that appends another command (e.g. a ``curl`` or package install) would silently inherit the token.

Since this is a GitHub token -- even if short-lived -- I would prefer to limit the scope to only the ``quarto install`` invocation. The simplest approach that keeps the template clean:

```
if [ -s /run/secrets/github_token ]; then export GH_TOKEN="$(cat /run/secrets/github_token)"; fi && \
{{ quarto_binary }} install tinytex --no-prompt --quiet && \
unset GH_TOKEN
```

Or, for tighter scoping via command-prefix form (no ``export`` at all):

```
if [ -s /run/secrets/github_token ]; then
  GH_TOKEN="$(cat /run/secrets/github_token)" {{ quarto_binary }} install tinytex --no-prompt --quiet;
else
  {{ quarto_binary }} install tinytex --no-prompt --quiet;
fi
```

The prefix form makes the scope explicit and self-documenting. The ``unset`` approach is simpler to template.

bschwedler (commented):
Review feedback assisted by the [critical-code-reviewer skill](https://github.com/posit-dev/skills/blob/main/posit-dev/critical-code-reviewer/SKILL.md).

posit-bakery/posit_bakery/image/bake/bake.py
ianpittwood: I simplified this in #486

posit-bakery/posit_bakery/config/templating/macros/quarto.j2
ianpittwood: I made it inline now. Check it out, it should be sufficient.

PR #488, opened 2026-04-24
Migrate docs to great-docs

bschwedler:
## Summary

- Use great-docs to generate API reference pages from the package source, checked into `docs/reference/`
- Keep the existing Quarto project with posit-docs theme — great-docs is a pre-build step, not the build system
- `just docs-preview` and `just docs-render` regenerate reference pages before rendering

## What changed

- `posit-bakery/docs/reference/` — generated API reference pages (ci, config, const, error, image, plugins, registry_management, services, util)
- `posit-bakery/docs/_quarto.yml` — Reference navbar link and sidebar section
- `posit-bakery/great-docs.yml` — config for reproducible reference generation
- `posit-bakery/justfile` — `docs-reference` recipe; `docs-preview`/`docs-render` depend on it
- `posit-bakery/pyproject.toml` — `[project.urls]`
- `posit-bakery/.gitignore` — `great-docs/`

### DIFF
--- a/posit-bakery/great-docs.yml
+++ b/posit-bakery/great-docs.yml
@@ -1,0 +1,57 @@
+display_name: Posit Bakery
+changelog:
+  enabled: false
+hero: false
+markdown_pages: false
+
+source:
+  branch: main
+
+inline_methods: true
+
+auto_include:
+  - config
+
+cli:
+  enabled: false
+
+exclude:
+  - cli
+  - main
+  - app
+
+reference:
+  sections:
+    - title: Configuration
+      desc: Models for bakery.yaml
+      contents:
+        - config.BakeryConfig
+        - config.BakeryConfigDocument
+        - config.Image
+        - config.ImageVariant
+        - config.ImageVersion
+        - config.ImageVersionOS
+        - config.BaseRegistry
+        - config.Registry
+        - config.Repository
+        - config.TagPattern
+    - title: CI
+      desc: CI pipeline operations
+      contents:
+        - ci.ci_matrix
+        - ci.ci_merge
+        - ci.resolve_metadata_globs
+        - ci.ImageTarget
+    - title: Image
+      desc: Image build targets and plans
+      contents:
+        - image.BakePlan
+        - image.ImageTarget
+        - image.ImageBuildStrategy
+    - title: Errors
+      desc: Exception hierarchy
+      contents:
+        - error.BakeryError
+        - error.BakeryToolError
+        - error.BakeryToolRuntimeError
+        - error.BakeryBuildErrorGroup
--- a/posit-bakery/great-docs.yml
+++ b/posit-bakery/great-docs.yml
@@ -1,0 +1,52 @@
+display_name: Posit Bakery
+changelog:
+  enabled: false
+hero: false
+markdown_pages: false
+
+source:
+  branch: main
+
+inline_methods: true
+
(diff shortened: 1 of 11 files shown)


posit-bakery/pyproject.toml
```diff
@@ -66,6 +70,9 @@ dev = [
     "pytest-mock~=3.14.1",
     "pytest-xdist~=3.8.0",
 ]
+docs = [
+    "great-docs @ file:///home/bschwedler/repos/posit-dev/great-docs ; python_version >= '3.11'",
```
ianpittwood: I see you have a fork PR'd into great-docs, but so we don't forget:

```suggestion
    "great-docs>=0.9,<1.0 ; python_version >= '3.11'",
```

For now I am using
```toml
docs = [
    "great-docs @ git+https://github.com/bschwedler/great-docs.git@inline-methods-config ; python_version >= '3.11'",
]
```

posit-bakery/great-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+source:
+  branch: main
+
+inline_methods: true
+
+auto_include:
+  - config
+
+cli:
+  enabled: false
```
ianpittwood: Do we want to consider enabling this? I know it may not work with all our callbacks/wrappers, but it would be a handy reference.

posit-bakery/great-docs.yml
```diff
@@ -0,0 +1,57 @@
+display_name: Posit Bakery
+changelog:
+  enabled: false
```
ianpittwood, line 3: This could be useful in the future I bet.

posit-bakery/great-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+        - image.BakePlan
+        - image.ImageTarget
+        - image.ImageBuildStrategy
+    - title: Errors
+      desc: Exception hierarchy
+      contents:
+        - error.BakeryError
+        - error.BakeryToolError
+        - error.BakeryToolRuntimeError
+        - error.BakeryBuildErrorGroup
```
ianpittwood: We're missing a handful of errors here. Were they left out for a reason? We probably should go and clean them up later anyways.

posit-bakery/great-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+cli:
+  enabled: false
+
+exclude:
+  - cli
+  - main
+  - app
+
+reference:
+  sections:
```
ianpittwood, line 26: Should we expose a section for `registry_management`?

posit-bakery/great-docs.yml
bschwedler: I would like to move to a more human-readable changelog before doing this.

posit-bakery/great-docs.yml
bschwedler: Fixed by using the top-level `error` only.

posit-bakery/great-docs.yml
bschwedler: Done. This is explicit for now because the `__init__.py` is empty. It's OK right now, but it may cause drift pain later.

ianpittwood (changes requested):
I think at least the Plugin protocol should be exposed. Otherwise, LGTM.

ianpittwood (approved):

PR #494, opened 2026-04-28
Add Ubuntu 26.04 support to bakery

bschwedler:
## Summary

- Add Ubuntu 26.04 (resolute) to ``SUPPORTED_OS``
- Add ``codename`` parameter to ``apt.setup()`` / ``apt.run_setup()`` macro chain so templates can configure the Posit Cloudsmith apt repo using a different codename when packages are not yet published for the current OS

The ``codename`` parameter passes the value as an environment variable to the Cloudsmith setup script, which natively supports overriding OS detection via ``codename=<value>``.

## Test plan

- [x] Existing bakery tests pass
- [ ] Verify rendered Containerfiles use the correct codename when ``codename="noble"`` is passed
- [ ] Build a dev image with Ubuntu 26.04 base after merging downstream PRs


posit-bakery/posit_bakery/config/templating/macros/apt.j2
```diff
@@ -3,14 +3,26 @@
...
-{# Setup command for Posit Cloudsmith apt repositories #}
-{% macro setup_posit_cloudsmith() -%}
+{# Setup command for Posit Cloudsmith apt repositories.
+
+:param codename_override: Override the OS codename used by the setup script. Useful when
+    Posit packages are not yet published for the current OS (e.g., using noble packages on
+    resolute). The original codename is restored after setup.
+#}
+{% macro setup_posit_cloudsmith(codename_override = None) -%}
+{%- if codename_override -%}
```
bschwedler: I was uncertain if we wanted to keep this `codename_override` to be very explicit about the naming. `codename` would be cleaner, but doesn't suggest that it is an override.

bschwedler:
This relies on the instructions here: https://cloudsmith.io/~posit/repos/pro/setup/#formats-deb

```bash
curl -1sLf \
  'https://dl.posit.co/public/pro/setup.deb.sh' \
  | sudo -E distro=DISTRO codename=CODENAME arch=ARCH component=COMPONENT bash
```

posit-bakery/posit_bakery/config/templating/macros/apt.j2
```diff
@@ -3,14 +3,24 @@
...
+{# Setup command for Posit Cloudsmith apt repositories.
+
+:param codename: Override the OS codename used by the Cloudsmith setup script. Useful when
+    Posit packages are not yet published for the current OS (e.g., using noble packages on
+    resolute). When set, the codename is passed as an environment variable to the script
+    per the Cloudsmith docs.
+#}
+{% macro setup_posit_cloudsmith(codename = None) -%}
+{%- if codename -%}
+codename={{ codename }} bash -c "$(curl -1fsSL 'https://dl.posit.co/public/pro/setup.deb.sh')"
```
ianpittwood, line 15: nit: quote, will also require test updates

```suggestion
codename="{{ codename }}" bash -c "$(curl -1fsSL 'https://dl.posit.co/public/pro/setup.deb.sh')"
```

ianpittwood (approved):

PR #496, opened 2026-04-28
Strip v prefix from image-version input

bschwedler:
## Summary

Fixes the empty-matrix failure seen on https://github.com/posit-dev/images-connect/actions/runs/25058571626 — the first real dispatch from a product repo after #480 landed.

Product repos dispatch with a git-describe version like `v2026.03.0-473-g072bb6fd1f`. Bakery version names never carry the `v` prefix, so `--image-version` matched nothing and the matrix came back empty.

## Changes

- Strip leading `v` from `IMAGE_VERSION` with `${IMAGE_VERSION#v}` before building CLI args (both `bakery-build-native.yml` and `bakery-build.yml`).
- Separate `bakery ci matrix` from the `echo` that writes `$GITHUB_OUTPUT` so `set -e` can catch non-zero exits. Previously the exit code was swallowed inside a command substitution within `echo`.

## Test plan

- [ ] Re-dispatch Connect dev build — matrix should populate correctly.


.github/workflows/bakery-build-native.yml
```diff
@@ -119,10 +119,12 @@ jobs:
           DEV_STREAM: ${{ inputs.dev-stream }}
           CONTEXT: ${{ inputs.context }}
         run: |
+          IMAGE_VERSION="${IMAGE_VERSION#v}"
```
ianpittwood, line 122: We may want to consider adding a callout in CI.md

ianpittwood (approved):

Issue #484, opened 2026-04-24
Push image versions in reverse order


ianpittwood:
Docker Hub always orders tags based on the most recent push. As our builds run in parallel, the order in which we pus is always random. Sometimes older versions may show as the default image to pull regardless of which image actually has the latest tag. We should fix this as it could cause a poor user experience when discovering our images in Docker Hub.

PR #501, opened 2026-04-29
Implement latest tag for image matrix builds

### LINKED ISSUE #498: Implement a `latest` tag behavior for image matrix builds
Related to #484 

If we want to resolve the same issue for the push order of matrix images, we will need to add a concept of `latest` and be able to resolve the images in a versioned order.

ianpittwood:
Closes #498.

## Summary

- Adds `ImageMatrix.latest_combination` that resolves the cartesian-product row at the maximum version of every dependency and list-typed value, returning `None` (with a warning) for non-version-parseable axes.
- `ImageMatrix.to_image_versions()` now sets `latest=True` on the single matching row via a new `_matches_latest` static helper.
- `default_matrix_tag_patterns()` gains the four `LATEST`-filtered patterns from `default_tag_patterns()`, so the existing `ImageTarget.tag_patterns` filter machinery emits `latest`, `<os>`, `<variant>`, and `<os>-<variant>` on the right rows.

No changes to `ImageTarget`, README push, CI workflows, or other consumers — they already handle `latest=True` correctly.

This unblocks #484 (push-order in CI) by giving matrix images a meaningful `latest` signal.

## Test Plan

- [x] Unit tests cover happy path, `None`-on-unparseable, short-circuit, scalar/list value distinction, original-string preservation, `latest=True` propagation, default-pattern parity (1439 tests pass)
- [x] Integration test asserts `ImageTarget.tag_suffixes` emits the bare `latest` tag for the latest matrix row + primary OS + primary variant
- [x] Empirical validation against `images-connect/connect-content`:
  - Latest combination (R4.6.0/python3.14.4/base/ubuntu-24.04) → emits `latest`, `base`, `ubuntu-24.04`, `ubuntu-24.04-base` (+ existing matrix tags)
  - Latest deps + non-primary variant (pro) → emits `pro`, `ubuntu-24.04-pro` only
  - Latest deps + non-primary OS (22.04) → emits `ubuntu-22.04`, `ubuntu-22.04-base` only
  - Older deps (R4.3.3/python3.11.15) → no \`latest\`-family tags
- [x] Documentation updated (`posit-bakery/docs/configuration.qmd` ImageMatrix section)

### DIFF
--- a/posit-bakery/posit_bakery/config/image/matrix.py
+++ b/posit-bakery/posit_bakery/config/image/matrix.py
@@ -415,6 +415,13 @@ class ImageMatrix(BakeryPathMixin, BakeryYAMLModel):
 
         selected: dict[str, str] = {}
         for axis_key, candidates in axes:
+            if not candidates:
+                log.warning(
+                    f"Image matrix '{self.namePattern}': cannot determine latest because axis "
+                    f"'{axis_key}' has no candidate versions. "
+                    f"No 'latest'-family tags will be emitted for this matrix."
+                )
+                return None
             parsed: list[tuple[DependencyVersion, str]] = []
             for candidate in candidates:
                 try:
--- a/posit-bakery/posit_bakery/config/image/matrix.py
+++ b/posit-bakery/posit_bakery/config/image/matrix.py
@@ -693,8 +693,10 @@ class ImageMatrix(BakeryPathMixin, BakeryYAMLModel):
         """
         image_versions = []
 
+        latest_pick = self.latest_combination
         products = self._cartesian_product(self.resolved_dependencies, self.values)
         for product in products:
+            is_latest = latest_pick is not None and self._matches_latest(product, latest_pick)
             image_version = ImageVersion(
                 parent=self.parent,
                 name=self._render_name_pattern(self.namePattern, product["dependencies"], product["values"]),
@@ -705,8 +707,26 @@ class ImageMatrix(BakeryPathMixin, BakeryYAMLModel):
                 dependencies=product["dependencies"],
                 values=product["values"],
                 isMatrixVersion=True,
+                latest=is_latest,
                 buildTarget=self.buildTarget,
             )
             image_versions.append(image_version)
 
         return image_versions
+
+    @staticmethod
+    def _matches_latest(product: dict[str, list | dict], latest_pick: dict[str, str]) -> bool:
+        """Return True iff every dependency and list-typed value in ``product`` equals
+        the corresponding entry in ``latest_pick`` (by original-string equality).
+
+        Scalar values are not compared — they are constant across all rows.
+        """
+        for dep in product["dependencies"]:
+            expected = latest_pick.get(dep.dependency)
+            if expected is None or dep.versions[0] != expected:
+                return False
+        for key, value in product["values"].items():
+            axis_key = f"value:{key}"
+            if axis_key in latest_pick and value != latest_pick[axis_key]:
+                return False
+        return True
--- a/posit-bakery/posit_bakery/config/image/matrix.py
+++ b/posit-bakery/posit_bakery/config/image/matrix.py
@@ -412,8 +412,7 @@ class ImageMatrix(BakeryPathMixin, BakeryYAMLModel):
                         f"No 'latest'-family tags will be emitted for this matrix."
                     )
                     return None
-            max_version, max_original = max(parsed, key=lambda pair: pair[0])
-            selected[axis_key] = max_original
+            selected[axis_key] = max(parsed, key=lambda pair: pair[0])[1]
 
         return selected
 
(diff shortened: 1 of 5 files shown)


posit-bakery/posit_bakery/config/image/matrix.py
```diff
@@ -374,6 +375,47 @@ def resolved_dependencies(self) -> list[DependencyVersions]:
...
+
+        selected: dict[str, str] = {}
+        for axis_key, candidates in axes:
+            parsed: list[tuple[DependencyVersion, str]] = []
+            for candidate in candidates:
+                try:
+                    parsed.append((DependencyVersion(candidate), candidate))
+                except Exception as e:
+                    # packaging raises InvalidVersion (a ValueError subclass); we keep this
+                    # broad to also catch any defensive parsing errors from DependencyVersion.
```
bschwedler: I would like to see this narrowed down if possible, even if we have a list of exception types that we will catch

If this encounters an error that is not related to parsing, the warning message could be a red herring.

posit-bakery/posit_bakery/config/image/matrix.py
```diff
@@ -651,8 +693,10 @@ def to_image_versions(self) -> list[ImageVersion]:
         """
         image_versions = []
 
+        latest_pick = self.latest_combination
         products = self._cartesian_product(self.resolved_dependencies, self.values)
```
bschwedler: These lines end up calling `self.resolved_dependencies` twice.  Can we add an internal class attr to `resolved_dependencies` that we short-circuit with a return if it is already set?

posit-bakery/posit_bakery/config/image/matrix.py
```diff
@@ -374,6 +375,47 @@ def resolved_dependencies(self) -> list[DependencyVersions]:
...
+                except Exception as e:
+                    # packaging raises InvalidVersion (a ValueError subclass); we keep this
+                    # broad to also catch any defensive parsing errors from DependencyVersion.
+                    log.warning(
+                        f"Image matrix '{self.namePattern}': cannot determine latest because axis "
+                        f"'{axis_key}' has unparseable version '{candidate}' ({e}). "
+                        f"No 'latest'-family tags will be emitted for this matrix."
+                    )
+                    return None
+            selected[axis_key] = max(parsed, key=lambda pair: pair[0])[1]
```
bschwedler, line 462: Can we gate this on `parsed` having a length > 0?

I see you noted this failure case in the PR description.

posit-bakery/posit_bakery/config/tag.py
```diff
@@ -161,4 +160,20 @@ def default_matrix_tag_patterns() -> list[TagPattern]:
...
+            only=[TagPatternFilter.LATEST, TagPatternFilter.PRIMARY_VARIANT],
+        ),
+        TagPattern(
+            patterns=["{{ Variant }}"],
+            only=[TagPatternFilter.LATEST, TagPatternFilter.PRIMARY_OS],
+        ),
+        TagPattern(
+            patterns=["latest"],
+            only=[TagPatternFilter.LATEST, TagPatternFilter.PRIMARY_OS, TagPatternFilter.PRIMARY_VARIANT],
+        ),
```
bschwedler: This duplicates tags that are in the `default_tag_patterns`.  I think it would be useful to break out some of the common lines between versioned images and matrix images into helper functions to help with consistency: 

It could end up looking something like
```python
  def default_tag_patterns() -> list[TagPattern]:                                                                                                                                                                                                                               
      return [                                                     
          *_version_os_variant_patterns(["{{ Version }}", "{{ Version | stripMetadata }}"]),
          *_latest_tag_patterns(),                                                                                                                                                                                                                                              
          *_version_os_variant_patterns(["{{ Stream }}"]),
      ]                                                                                                                                                                                                                                                                         
                                                                   
                                                                                                                                                                                                                                                                                
  def default_matrix_tag_patterns() -> list[TagPattern]:
      return [                                                                                                                                                                                                                                                                  
          *_version_os_variant_patterns(["{{ Version }}"]),        
          *_latest_tag_patterns(),
      ]
```

bschwedler (changes requested):

posit-bakery/posit_bakery/config/image/matrix.py
ianpittwood: I added a catch for this issue.

posit-bakery/posit_bakery/config/image/matrix.py
ianpittwood: Separated into a parsing error and an unexpected error catch.

posit-bakery/posit_bakery/config/image/matrix.py
ianpittwood: I changed it to an `@cached_property` decorated function

posit-bakery/posit_bakery/config/tag.py
ianpittwood: I took this halfway. I don't really like the arg passing with your example so I only did fully shared patterns.

bschwedler (approved):

Issue #330, opened 2026-01-21
Limit when we retrieve dependency versions


bschwedler:
Currently, this log message is repeated multiple times:

```log
           DEBUG    Initializing backend: filesystem bakery_cache                                         __init__.py:66
           DEBUG    Initialized FileDict with serializer: SerializerPipeline(name=json, n_stages=2)          base.py:322
           DEBUG    Initialized SQLiteDict with serializer: None                                             base.py:322
           DEBUG    Opening connection to /tmp/bakery_cache/redirects.sqlite:redirects                     sqlite.py:252
           DEBUG    Cache directives from request headers: CacheDirectives()                              actions.py:103
           DEBUG    Pre-read cache checks: Passed                                                         actions.py:368
           DEBUG    Post-read cache actions: CacheActions(expire_after=3600)                              actions.py:217
```

While the retrieval of versions from live URLs is cached locally, we are doing this more than we need to.

Limit the retrieval to:
- Once per `bakery` invocation
- Only when resolving `DependencyConstraints`

PR #503, opened 2026-04-29
Defer dev version stream resolution to load time

bschwedler:
## Summary

- Move stream resolution (fetching version/URL from product CDN) out of Pydantic model validation and into ``load_dev_versions()``, which only runs when dev versions are actually needed
- Per-OS resolution failures now exclude the failing OS with a warning instead of crashing config loading
- Production builds (``--dev-versions exclude``) are no longer affected by stream availability

## Context

The ``add_os_url`` model validator on ``BaseImageDevelopmentVersion`` ran during ``bakery.yaml`` parsing, eagerly fetching stream data for every OS in every dev version. If any platform was missing from the stream endpoint (e.g. Connect has no Ubuntu 26.04 dailies yet), config loading failed with a Pydantic ``ValidationError`` -- blocking **all** builds including production.

This caused the CI failure on posit-dev/images-connect#73 (Ubuntu 26.04 support).

Partially addresses #330 (limit when we retrieve dependency versions) by eliminating stream fetches during config loading.

## Test plan

- [x] 987 config/CI tests pass
- [x] New tests for ``_resolve_os_urls()``: URL population, partial OS failure, total failure
- [ ] Re-run posit-dev/images-connect#73 CI after merge

### DIFF
--- a/posit-bakery/posit_bakery/config/image/dev_version/stream.py
+++ b/posit-bakery/posit_bakery/config/image/dev_version/stream.py
@@ -19,7 +19,15 @@ class ImageDevelopmentVersionFromProductStream(BaseImageDevelopmentVersion):
     sourceType: Literal["stream"] = "stream"
     product: Annotated[ProductEnum, Field(description="The ID of the product stream to use for this image version.")]
     stream: Annotated[ReleaseStreamEnum, Field(description="The release stream to use for this image version.")]
-    _resolved_version: str | None = None
+    resolved_version: Annotated[
+        str | None,
+        Field(
+            exclude=True,
+            default=None,
+            description="Cached version from the last _resolve_os_urls() call. Avoids a redundant stream fetch in "
+            "get_version().",
+        ),
+    ]
 
     def get_primary_os(self) -> ImageVersionOS:
         """Retrieve the primary OS from the parent image if available.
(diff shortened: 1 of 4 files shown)


posit-bakery/posit_bakery/config/image/dev_version/stream.py
```diff
@@ -30,11 +35,15 @@ def get_primary_os(self) -> ImageVersionOS:
     def get_version(self) -> str:
         """Retrieve the version from the specified product stream.
 
+        If _resolve_os_urls() has already been called, returns the cached
+        version. Otherwise fetches it from the primary OS stream.
+
         :return: The version string from the product stream.
         """
+        if self._resolved_version is not None:
+            return self._resolved_version
```
ianpittwood: This should already be cached by requests IIRC. Is this just to avoid recalling that code?

posit-bakery/posit_bakery/config/image/dev_version/stream.py
```diff
@@ -1,20 +1,25 @@
...
+log = logging.getLogger(__name__)
+
 
 class ImageDevelopmentVersionFromProductStream(BaseImageDevelopmentVersion):
     """Image development version sourced from a product stream."""
 
     sourceType: Literal["stream"] = "stream"
     product: Annotated[ProductEnum, Field(description="The ID of the product stream to use for this image version.")]
     stream: Annotated[ReleaseStreamEnum, Field(description="The release stream to use for this image version.")]
+    _resolved_version: str | None = None
```
ianpittwood: Might be worth making the `Field` definition explicit here to make it clear this is a private, non-serializable attribute. Also it would then match the `Annotated` patterns we commonly use.

ianpittwood (approved):
I think this makes sense, but for whatever reason my brain is having a hard time with it 😂

posit-bakery/posit_bakery/config/image/dev_version/stream.py
bschwedler: Done in 1c47c1b -- made it an explicit `Annotated[..., Field(exclude=True)]` field.

PR #505, opened 2026-04-29
Push image versions in deterministic order to fix Docker Hub display

### LINKED ISSUE #484: Push image versions in reverse order

ianpittwood:
## Summary
- Adds `ImageTarget.push_sort_key`, a sort tuple driving Docker Hub display order: `(image_name, is_latest, ParsedVersion key, primary_score, version.name, variant.name, os.name)`. Latest pushes last; ties broken by primary OS+variant; multi-image runs grouped per image.
- Sorts targets at the top of `OrasPlugin.execute` and emits `log.info("ORAS merge order: ...")` so CI logs make the chosen order observable.
- Collapses the `bakery-build-native.yml` `merge` matrix into a single ordered job; drops the now-unused `versions-matrix` output and the `Images by Version` step. Build/test fan-out is unchanged.

Closes #484.

## Dependencies
Builds on (must land first or merge together):
- #502 — `ParsedVersion` value type and `version_sort_key` helper
- #501 — `latest=True` on the canonical matrix combination

Both are merged into this branch already, so the diff against `main` will collapse to just the ordered-push commits once they land.

## Trade-off
Wall-time on the `merge` job grows because pushes are now sequential rather than parallel:

| Workflow | Old | New | Δ |
|---|---|---|---|
| Connect production (~5 versions) | ~2 min | ~4–5 min | +2–3 min |
| `connect-content` matrix (~12 combos × variants) | ~2 min | ~7–10 min | +5–8 min |
| PR builds (`push: false`, dry-run) | ~2 min | ~2 min | negligible |

Total runner-minutes drop because per-runner setup (Docker, buildx, ORAS) only runs once.

## Test plan
- [x] `just test` — 1487 passed, 0 failures, 84% coverage
- [x] `TestPushSortKey` (7 new tests): latest-last, primary-within-version, version ordering, semver §11 prerelease (`daily < dev < release`), matrix non-latest tiebreakers, matrix latest sorts last, multi-image grouping
- [x] `OrasPlugin.execute` ordering test: scrambled input → sorted output, exactly one `ORAS merge order:` log line
- [x] `pre-commit` actionlint hook on `bakery-build-native.yml` — clean
- [ ] First post-merge production run: confirm `ORAS merge order:` line in CI log shows the expected sequence
- [ ] First post-merge production run: visit `hub.docker.com/r/posit/connect` and confirm the `latest`-aliased tag is the most-recent push
- [ ] Same check for `posit/connect-content` after the next content workflow run (uses #501's `latest_combination`)

### DIFF
--- a/.github/workflows/bakery-build-native.yml
+++ b/.github/workflows/bakery-build-native.yml
@@ -99,7 +99,6 @@ jobs:
       contents: read
     outputs:
       platform-matrix: ${{ steps.images-by-platform.outputs.platform_matrix }}
-      versions-matrix: ${{ steps.images-by-version.outputs.versions_matrix }}
 
     steps:
       - name: Checkout
@@ -125,21 +124,6 @@ jobs:
           [[ -n "$DEV_STREAM" ]] && ARGS+=(--dev-stream "$DEV_STREAM")
           result=$(bakery ci matrix "${ARGS[@]}")
           echo "platform_matrix=$(echo "$result" | jq --compact-output .)" >> "$GITHUB_OUTPUT"
-      - name: Images by Version
-        id: images-by-version
-        env:
-          DEV_VERSIONS: ${{ inputs.dev-versions }}
-          MATRIX_VERSIONS: ${{ inputs.matrix-versions }}
-          IMAGE_VERSION: ${{ inputs.image-version }}
-          DEV_STREAM: ${{ inputs.dev-stream }}
-          CONTEXT: ${{ inputs.context }}
-        run: |
-          IMAGE_VERSION="${IMAGE_VERSION#v}"
-          ARGS=(--quiet --dev-versions "$DEV_VERSIONS" --matrix-versions "$MATRIX_VERSIONS" --exclude platform --context "$CONTEXT")
-          [[ -n "$IMAGE_VERSION" ]] && ARGS+=(--image-version "$IMAGE_VERSION")
-          [[ -n "$DEV_STREAM" ]] && ARGS+=(--dev-stream "$DEV_STREAM")
-          result=$(bakery ci matrix "${ARGS[@]}")
-          echo "versions_matrix=$(echo "$result" | jq --compact-output .)" >> "$GITHUB_OUTPUT"
 
   build-test:
     name: "Build/Test ${{ matrix.img.image }}:${{ matrix.img.version }} (${{ matrix.img.platform }})"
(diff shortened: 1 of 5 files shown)


posit-bakery/posit_bakery/image/image_target.py
```diff
@@ -314,6 +315,31 @@ def is_primary_variant(self) -> bool:
             return True
         return self.image_variant.primary
 
+    @property
+    def push_sort_key(self) -> tuple:
```
bschwedler: Can we type this tuple, or make it a named tuple?
```suggestion
    def push_sort_key(self) -> tuple[str, bool, ParsedVersion, int, str, str, str]:
```

```python
class PushSortKey(NamedTuple):
      image_name: str
      is_latest: bool
      version: ParsedVersion
      primary_score: int
      version_name: str
      variant_name: str
      os_name: str
```

.github/workflows/bakery-build-native.yml
```diff
@@ -357,7 +335,7 @@ jobs:
       - name: Download Metadata
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
         with:
-          pattern: "${{ matrix.img.image }}-${{ matrix.img.version }}-*-metadata"
+          pattern: "*-metadata"
```
bschwedler, line 338: Should we make this more specific so we don't match generic job names like `my-metadata`, or do we actually need it to be this generic?
```suggestion
          pattern: "*-*-*-metadata"
```

ianpittwood: I guess what is your concern? As I understand, artifacts are isolated to the job they are created by. We shouldn't be able to do something like accidentally pull an artifact from the session images in the product image CI. Nor will we be able to pull a previous workflow iteration's artifacts.

posit-bakery/posit_bakery/image/image_target.py
ianpittwood: Fixed

.github/workflows/bakery-build-native.yml
bschwedler: Thanks for the explanation. That makes sense now.

bschwedler (approved):

PR #510, opened 2026-04-30
Use segment-aware prefix matching for --image-version

bschwedler:
## Summary

- Add `version_matches()` to `config.py` for segment-aware prefix matching on `--image-version`
- Replace `re.search()` version filtering in `build`, `run`, `get tags`, `update files`, and `ci matrix`
- Remove `re.escape()` at call sites and unused `import re` from CLI modules
- Falls back to dot-separated segment comparison for non-semver strings (e.g. `R4.5.3-python3.14.3`)
- Update help text and CLAUDE.md to reflect prefix matching behavior

The previous regex approach had latent bugs: `re.search("2026.1", "2026.11.0")` matched across segment boundaries. Prefix matching compares parsed version segments instead.

Based on #502 to reuse `ParsedVersion` for structured comparison.

Fixes the dispatch failure at https://github.com/posit-dev/images-connect/actions/runs/25181615909 where `--image-version 2026.05` didn't match `2026.05.0-dev+15-gSHA`.


ianpittwood (changes requested):
Can you apply the prefix matching on the option to all commands that use it? I'd love to have that feature.

bschwedler:
> Can you apply the prefix matching on the option to all commands that use it? I'd love to have that feature.

@ianpittwood Done.

I was also thinking: Do you think we should add a `--regex` flag or similar to toggle whether the other matchers use regex or not? This would be a follow-up PR, but I sometimes forget that `--image-name connect` will match all the images until I anchor with the regex.

ianpittwood (approved):

ianpittwood:
> > Can you apply the prefix matching on the option to all commands that use it? I'd love to have that feature.
> 
> @ianpittwood Done.
> 
> I was also thinking: Do you think we should add a `--regex` flag or similar to toggle whether the other matchers use regex or not? This would be a follow-up PR, but I sometimes forget that `--image-name connect` will match all the images until I anchor with the regex.

You could make an issue for it. We should figure out a better way of handling a string versus a regex pattern, but it could be its own change.

bschwedler:
Created: #519

PR #514, opened 2026-05-06
feat(product-release): regenerate Supported tags sections in image READMEs

ianpittwood:
## Summary

- Adds a new step in `product-release.yml` after `Update READMEs` that regenerates the `## Supported tags` bullet list in each image's README.
- Per image, skips silently if the README is missing or lacks the section, so this is a no-op for repos that haven't adopted the section.
- Tags are pulled from `bakery build --plan` and curated to match the existing hand-maintained format (drop `-std` redundant suffixes; for non-latest versions, drop verbose-version+OS combinations and emit only the std+primary-OS Containerfile).

## Curation rules

- **Containerfile filter** — all (variant × OS) combos for the latest version; only `std` + primary OS for older versions.
- **Tag filter** — drop tags ending in `-std`; for non-latest versions, also drop `<verbose-version>-ubuntu-*` tags (keep only `<display-version>-ubuntu-*`).
- **Sort** — newer version first, primary OS first, std before min.

The bullet block under the heading is replaced; trailing prose (e.g. "For a full list of available tags...") is preserved because it sits past a blank line.

## Verification

- Dry-run against `images-package-manager` produced output identical to the manually curated section.
- Re-running on the resulting README is a no-op (idempotent).

## Test plan

- [ ] Trigger a release in a repo with a `Supported tags` section (e.g. `images-package-manager`) and confirm the bullet list updates correctly in the resulting release PR.
- [ ] Trigger a release in a repo without a `Supported tags` section and confirm the step skips with a notice log.
- [ ] Trigger a release in a repo without a README.md for a given image and confirm the step skips with a notice log.

## Caveats

- Primary OS and latest version are inferred from the tag set (Containerfile owning the bare `latest` floating tag → latest; Containerfile owning the bare `<display-version>` / `<display-version>-min` tag → primary OS). Stable today; would silently mis-bucket if Bakery's floating-tag patterns change. A follow-up to expose pattern metadata from Bakery directly would remove this fragility.
- The heading match is exact (`^## Supported tags\s*\n\n` followed by a contiguous bullet block). Different casing, extra blank lines, or interleaved prose will skip the file.

### DIFF
--- a/.github/workflows/product-release.yml
+++ b/.github/workflows/product-release.yml
@@ -138,6 +138,174 @@ jobs:
               -exec sed -i "s/${OLD_EDITION}/${EDITION}/g" {} +
           fi
 
+      - name: Update Supported tags sections
+        env:
+          IMAGES: ${{ inputs.images }}
+        run: |
+          # Regenerate the "## Supported tags" bullet list in each image's
+          # README from the current bakery build plan. Skips images that lack
+          # a README.md or that don't have a Supported tags section.
+          #
+          # Curation rules (matching the curated section style):
+          # - For the latest version, emit a bullet for every (variant, OS)
+          #   Containerfile. For older versions, emit only the std + primary
+          #   OS Containerfile.
+          # - Within each bullet's tag set, drop tags ending in "-std" (they
+          #   are redundant with the bare form). For non-latest versions,
+          #   also drop "<verbose-version>-ubuntu-*" tags (keep only the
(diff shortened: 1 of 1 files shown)


bschwedler (commented):
I like this approach, but given the naive `sed` that is done in the previous step, I think the supported tags list would likely be mutated first, and I'd like to understand the interaction.

PR #548, opened 2026-05-20
Normalize rendered template output for pre-commit

bschwedler:
Bakery's Jinja-rendered files (Containerfiles, goss.yaml, scripts, and bakery.yaml itself) could ship with trailing whitespace or a missing trailing newline, failing `trailing-whitespace` and `end-of-file-fixer` pre-commit hooks in consuming repos. Two release PRs failed CI for exactly this reason:

- https://github.com/posit-dev/images-connect/pull/77
- https://github.com/posit-dev/images-connect/pull/87

The root cause was two compounding bugs in the renderer. First, `ImageVersion.render_files` and `ImageMatrix.render_files` tried to enable `trim_blocks` for Containerfile templates by passing it to `tpl.render(trim_blocks=True)` — but `trim_blocks` is a Jinja2 `Environment` option, not a `Template.render()` argument, so it was a silent no-op that just landed in the template context. Second, no rendered-template write path normalized whitespace except `BakeryConfig.write()`, which had a one-off `re.sub` added after the first failure. Everything else (Containerfiles via `version.py`/`matrix.py`, goss.yaml, scripts, scaffolded `bakery.yaml`) wrote raw Jinja output, so `{% for %} ... {% endfor %}` blocks left lines containing only their leading indent — exactly what PR #87 caught in `connect/2026.04/test/goss.yaml`.

`normalize_rendered_output()` in `posit_bakery/config/templating/render.py` strips trailing spaces and tabs from each line and ensures a single trailing newline, matching what the two pre-commit hooks produce. It's now applied at every rendered-file write site, and `BakeryConfig.write()`'s inline `re.sub` is replaced by it for consistency. The dead `render_kwargs["trim_blocks"] = True` branch is removed since it never took effect; the new normalize step covers the same intent more robustly.

Verified by re-rendering all images in `images-connect`, `images-package-manager`, and `images-workbench` against the patched bakery — full pre-commit suite passes on all three.

### DIFF
--- a/posit-bakery/posit_bakery/config/config.py
+++ b/posit-bakery/posit_bakery/config/config.py
@@ -25,7 +25,7 @@ from posit_bakery.config.registry import BaseRegistry
 from posit_bakery.config.repository import Repository
 from posit_bakery.config.shared import BakeryPathMixin, BakeryYAMLModel
 from posit_bakery.config.templating import TPL_CONTAINERFILE, TPL_BAKERY_CONFIG_YAML
-from posit_bakery.config.templating.render import jinja2_env
+from posit_bakery.config.templating.render import jinja2_env, normalize_rendered_output
 from posit_bakery.config.image.parsed_version import ParsedVersion
 from posit_bakery.config.image.posit_product.const import ReleaseStreamEnum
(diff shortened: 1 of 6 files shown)


ianpittwood (approved):
LGTM, thanks for fixing this!

PR #552, opened 2026-05-28
Cache dependency version fetches per invocation

### LINKED ISSUE #304: Dev version builds occasionally fail in native builder workflows
Development versions occasionally fail to build when using the native builder CI workflow if a newer dev version is created perfectly in between the build and merge jobs. The probable solution for this is to reuse the requests cache from the build job in the merge job.

### LINKED ISSUE #330: Limit when we retrieve dependency versions

bschwedler:
Closes https://github.com/posit-dev/images-shared/issues/330
Closes https://github.com/posit-dev/images-shared/issues/304

Memoize the requests-cache session factory and each dependency's `_fetch_versions` helper so backend init and JSON fetch+parse run exactly once per `bakery` invocation, instead of once per `DependencyConstraint` instance.

- `available_versions()` returns a copy of the memoized list in all four dependency files, so callers can't corrupt the shared cache.
- `setup-bakery` restores the dependency-version HTTP cache from a run-scoped GitHub Actions cache and raises its TTL for the life of a run, so every job in a workflow (matrix, build-test, merge) resolves the same versions instead of racing against upstream between build time and merge time.

### DIFF
--- a/posit-bakery/posit_bakery/config/dependencies/positron.py
+++ b/posit-bakery/posit_bakery/config/dependencies/positron.py
@@ -81,9 +81,12 @@ class PositronDependency(BakeryYAMLModel, abc.ABC):
     def available_versions(self) -> list[DependencyVersion]:
         """Return a list of available Positron versions.
 
+        Returns a shallow copy since ``_fetch_versions`` is memoized; callers
+        must not mutate the cached list shared across all instances.
+
         :return: A sorted list of available Positron versions.
         """
-        return self._fetch_versions(self.prerelease)
+        return list(self._fetch_versions(self.prerelease))
 
 
 class PositronDependencyVersions(DependencyVersions, PositronDependency):
--- a/posit-bakery/posit_bakery/config/dependencies/positron.py
+++ b/posit-bakery/posit_bakery/config/dependencies/positron.py
@@ -1,4 +1,5 @@
 import abc
+import copy
 from functools import cache
 from typing import Annotated, Literal, ClassVar
 
@@ -81,12 +82,13 @@ class PositronDependency(BakeryYAMLModel, abc.ABC):
     def available_versions(self) -> list[DependencyVersion]:
         """Return a list of available Positron versions.
 
-        Returns a shallow copy since ``_fetch_versions`` is memoized; callers
-        must not mutate the cached list shared across all instances.
+        Returns a deep copy since ``_fetch_versions`` is memoized; callers
+        can freely mutate the returned list or its elements without
+        corrupting the shared cache.
 
         :return: A sorted list of available Positron versions.
         """
-        return list(self._fetch_versions(self.prerelease))
+        return copy.deepcopy(self._fetch_versions(self.prerelease))
 
 
 class PositronDependencyVersions(DependencyVersions, PositronDependency):
--- a/posit-bakery/posit_bakery/config/dependencies/positron.py
+++ b/posit-bakery/posit_bakery/config/dependencies/positron.py
@@ -1,4 +1,5 @@
 import abc
+from functools import cache
 from typing import Annotated, Literal, ClassVar
 
 from pydantic import ConfigDict, Field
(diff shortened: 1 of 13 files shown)


posit-bakery/posit_bakery/util.py
```diff
@@ -63,17 +64,19 @@ def auto_path() -> Path:
...
+    """
+    return CachedSession(
+        cache_name="bakery_cache",
+        expire_after=3600,
+        backend="filesystem",
+        use_temp=True,
+        allowable_methods=["GET"],
+        allowable_codes=[200],
+        stale_if_error=True,
+    )
```
ianpittwood, line 95: I have a crazy idea: what if we cached this somehow as part of `setup-bakery`? Give it some sort of reasonable TTL and then recall the artifact it creates across all jobs in a workflow. That way we can resolve #304 and edge cases where new versions can crop up between build time and merge time.

posit-bakery/posit_bakery/util.py
```diff
@@ -63,17 +64,19 @@ def auto_path() -> Path:
...
-        "expire_after": 3600,
-        "backend": "filesystem",
-        "use_temp": True,
-        "allowable_methods": ["GET"],
-        "allowable_codes": [200],
-        "stale_if_error": True,
-    }
-    session_kwargs.update(kwargs)
+@cache
+def cached_session() -> CachedSession:
```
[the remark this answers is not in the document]
ianpittwood: We shouldn't need to test core language functionality IMO. This isn't a complicated setup.

posit-bakery/posit_bakery/config/dependencies/positron.py
```diff
@@ -31,18 +32,21 @@ def releases_url(target_arch: str = _DEFAULT_ARCH) -> str:
         arch = _ARCH_MAP[target_arch]
         return POSITRON_RELEASES_URL_TEMPLATE.format(arch=arch)
 
-    def _fetch_versions(self) -> list[DependencyVersion]:
+    @staticmethod
+    @cache
```
ianpittwood, line 55: roborev noted that the cached result can be mutated. We may want to note that or find a way to always return a copy.

posit-bakery/test/config/conftest.py
```diff
@@ -92,6 +92,21 @@ def patch_testdata_response(url: str):
 
 @pytest.fixture(scope="function")
 def disable_requests_caching(mocker):
```
ianpittwood, line 100: Should this be autouse?

ianpittwood (approved):

posit-bakery/posit_bakery/util.py
bschwedler: I like this idea. Let me revisit this and see if there's a sneaky way to do this.

bschwedler: Implemented in `setup-bakery/action.yml`: restores `/tmp/bakery_cache` via `actions/cache`, keyed on `github.run_id`, and sets `BAKERY_CACHE_EXPIRE_AFTER=86400` so the restored cache does not expire mid-run. Every job in a workflow now resolves the same versions. Resolves #304.

posit-bakery/posit_bakery/config/dependencies/positron.py
bschwedler: Fixed in all four dependency files (positron/python/quarto/r), not just Positron: `available_versions()` now returns `list(self._fetch_versions(...))` instead of the cached list directly. No caller mutates it today, but this makes that a guarantee.

posit-bakery/test/config/conftest.py
bschwedler: Left as non-autouse — every test touching dependency fetches already requests `patch_requests_get`, which depends on this fixture. Can switch it if you want a safety net for tests that forget to request it.

bschwedler: Switched to autouse. It also caught three tests (`Image.test_create_matrix`, `test_create_matrix_existing_matrix_update_existing`, `BakeryConfig.test_target_generation_matrix`) that were silently making real HTTP calls to live dependency endpoints instead of using mocked data — fixed by requesting `patch_requests_get` like the rest of the suite.

setup-bakery/action.yml
```diff
@@ -28,3 +47,42 @@ runs:
...
+
+    - name: Restore dependency version cache
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
+      # No artifact yet (first job to run this invocation) is expected, not
+      # fatal: continue with an empty /tmp/bakery_cache and let it populate
+      # from upstream as normal. Tradeoff: this also swallows a real failure
+      # (bad name, permissions) identically to a cold cache, and the job
+      # that saves the cache will always show this step as failed-but
+      # -continued on its own restore attempt, since nothing exists yet.
+      continue-on-error: true
```
ianpittwood, line 93: It would probably be better to do this conditionally so only the writer can bypass restoring the dependency version cache, but composite actions make that difficult so we can also leave it as is.

ianpittwood (approved):
I think this approach looks good as is. Just one comment, but I'm not sure it's cleanly solvable.

setup-bakery/action.yml
bschwedler: Fixed in 22edd385: `setup-bakery` now takes an `is-writer` input that skips the restore step (`if: inputs.is-writer != 'true'`). Traced every `setup-bakery` call site — the writer role turned out to be fixed to the 3 `matrix` jobs in `bakery-build-native.yml`, `bakery-build.yml`, and `bakery-build-pr.yml`, all within this repo, so no cross-repo plumbing was needed. Set `is-writer: "true"` on those three; everything else keeps the default.

PR #555, opened 2026-05-28
feat: add SOCI indexing plugin and bakery ci publish orchestrator

### LINKED ISSUE #230: Build SOCI index file as part of pushing to repository
We need to build the SOCI index as part of the build manifest merge process for multi-platform builds.

CloudOps has told us that they will not be able to uses these images in the DataOps cluster unless they have SOCI support.

---
 
- https://github.com/awslabs/soci-snapshotter/blob/main/docs/soci-index-manifest-v2.md
- https://github.com/rstudio/rstudio-docker-products/pull/981

Can use `soci convert --all-platforms`
https://github.com/awslabs/soci-snapshotter/blob/main/docs/cli-usage.md#soci-convert

ianpittwood:
## Summary

- Adds a new `soci` builtin plugin that wraps `soci convert` / `soci push` and an in-module `ctr image pull` helper to materialize images in containerd. The plugin probes containerd namespaces (`default` → `moby`) and supports both non-standalone (default) and `--standalone` modes via `SociOptions` in `bakery.yaml`.
- **Standalone mode now performs a real OCI-layout round-trip.** `soci convert --standalone` only accepts a local OCI image layout (tar/directory) for source and destination — it never touches a registry. `SociConvertWorkflow._run_standalone` bridges the registry on both ends with oras and needs no containerd: `oras cp --to-oci-layout` (registry → layout) → `soci convert --standalone --format oci-dir` → push by digest with `oras cp --from-oci-layout` (the converted layout is written untagged) → scratch cleanup. `OrasCopy` gained `--from-oci-layout` / `--to-oci-layout` flags to support this.
- Splits `OrasMergeWorkflow` into three composable primitives (`OrasIndexCreateWorkflow`, `OrasIndexCopyWorkflow`, `OrasIndexCleanupWorkflow`) and rebuilds the existing merge workflow on top of them to preserve back-compat with `bakery oras merge`.
- Adds `bakery ci publish`, which composes oras index-create → optional SOCI convert (gated by `--enable-soci`) → oras index-copy → cleanup. `bakery ci merge` is preserved as a thin alias that delegates with `--no-enable-soci`.
- Adds a `setup-soci` composite action and wires an `enable-soci` workflow input + `Publish` step into `bakery-build-native.yml`. `bakery-build.yml` (QEMU) is explicitly deferred with an in-file note since it has no merge phase to interleave SOCI into.

The implementation follows the spec at `docs/superpowers/specs/2026-05-18-soci-indexing-design.md` (local-only) and the task-by-task plan at `docs/superpowers/plans/2026-05-18-soci-indexing.md` (local-only).

## Test plan

- [x] Full bakery test suite passes (no regressions in oras / soci / dgoss / ci suites)
- [x] \`bakery soci --help\` and \`bakery soci convert --help\` show the expected flags
- [x] \`bakery ci --help\` lists both \`merge\` (back-compat alias) and \`publish\` (new orchestrator)
- [x] Pre-commit hooks clean across all changed files
- [x] Standalone mode verified end-to-end: ran the real \`SociConvertWorkflow._run_standalone\` against a throwaway local registry — oras pulls the source into an OCI layout, \`soci convert --standalone\` produces a zTOC, and the \`-soci\` index is pushed back. Confirmed via \`oras manifest fetch\` that the published index contains the SOCI artifact (\`artifactType\` includes \`soci\`).
- [ ] Manual: build a small image, run \`bakery ci publish --enable-soci --dry-run\` and confirm the soci convert / push commands are produced as expected
- [ ] Manual: full run against a single image (e.g. \`connect-content-init\`) with \`enable-soci: true\` in a sibling repo's workflow — verify the published manifest via \`oras manifest fetch\` and confirm SOCI metadata is present
- [ ] Manual: confirm layer-digest stability across \`soci convert\` (spec verification item #3)
- [ ] Per-product opt-in (e.g. flipping \`tool: soci, enabled: true\` on a Connect image and \`enable-soci: true\` in \`images-connect\`'s production workflow) lands as a follow-up PR in that repo, not here.


ianpittwood:
Commit [`939587f`](https://github.com/posit-dev/images-shared/pull/555/commits/939587f811c57d34f3da6f9928964130c0cb63e9) MUST be dropped before merge.

.github/workflows/bakery-build-native.yml
```diff
@@ -52,6 +52,11 @@ on:
         default: true
         required: false
         type: boolean
+      enable-soci:
```
bschwedler: I will come back and do a full review, but I think this should just be `--soci` so that the flag would be `--no-soci` if/when we change the default to `true`
```suggestion
      soci:
```

jspiewak:
I suggest updating to SOCI 0.13.0 and using the new `--standalone` to avoid needing to rely upon `containerd`.
See https://github.com/rstudio/aws-main-services/pull/1932

ianpittwood:
> I suggest updating to SOCI 0.13.0 and using the new `--standalone` to avoid needing to rely upon `containerd`. See [rstudio/aws-main-services#1932](https://github.com/rstudio/aws-main-services/pull/1932)

I tried it out, but `--standalone` would require us to export the image each time which turns into a storage issue for us.

jspiewak:
> > I suggest updating to SOCI 0.13.0 and using the new `--standalone` to avoid needing to rely upon `containerd`. See [rstudio/aws-main-services#1932](https://github.com/rstudio/aws-main-services/pull/1932)
> 
> I tried it out, but `--standalone` would require us to export the image each time which turns into a storage issue for us.

Can you say more? I don't have enough of a picture to understand the optimization here.

ianpittwood:
> > > I suggest updating to SOCI 0.13.0 and using the new `--standalone` to avoid needing to rely upon `containerd`. See [rstudio/aws-main-services#1932](https://github.com/rstudio/aws-main-services/pull/1932)
> > 
> > 
> > I tried it out, but `--standalone` would require us to export the image each time which turns into a storage issue for us.
> 
> Can you say more? I don't have enough of a picture to understand the optimization here.

I guess we can since now I'm remembering we do a mostly incremental pull/merge/push process rather than parallel. I could see exporting the tarball costing us 30GB+ for some image constructions though. We would need to:

- Pull both platform builds. We build multiplatform images on native architectures to avoid the speed hit for emulation. It also allows us to test on native architectures.
- Merge the individual platforms together.
- Export the image to a tarball. This at least doubles the disk usage.
- Run the SOCI conversion in standalone, push it to temp.
- Push/copy images to final destinations.
- Clear local tarball and image copies.

When running on expanded runners, I think we get 100GB of disk storage. If that is the case, we should be able to do standalone mode. We just need to be very diligent about clean up after each image is processed.

jspiewak:
https://docs.github.com/en/actions/reference/runners/larger-runners#specifications-for-general-larger-runners

150 GB on the 4-core runners, 300 GB on the 8-core.

posit-bakery/posit_bakery/plugins/builtin/oras/oras.py
```diff
@@ -131,26 +131,185 @@ class OrasCopy(OrasCommand):
...
+            cmd.append("--descriptor")
+        cmd.append(self.reference)
+        return cmd
+
+
+class OrasIndexCreateResult(BaseModel):
+    """Result of an ORAS manifest-index-create phase."""
+
+    success: Annotated[bool, Field(description="Whether the create phase succeeded.")]
+    temp_ref: Annotated[str | None, Field(default=None, description="The temp ref of the created index.")]
```
bschwedler: `None` does not be to looked to pass in.

```suggestion
    temp_ref: Annotated[str, Field(description="The temp ref of the created index.")]
```

posit-bakery/docs/configuration.qmd
```diff
@@ -548,6 +548,46 @@ A GossOption represents options for `dgoss` tests against an image target.
...
 | `wait`<br/>*int*                   | The number of seconds to wait after container startup before running tests. Useful if there is a service that needs to complete its startup prior to testing. | `0`                                                                                                                | `30`             |
 
+#### SociOptions
+
+A SociOption configures [SOCI](https://github.com/awslabs/soci-snapshotter) (Seekable OCI) index generation for an image target. SOCI indexes let container runtimes lazily pull image layers, reducing startup latency for large images.
+
+SOCI options can be set in the `options` array of an [Image](#image) or an [ImageVariant](#imagevariant). When both are present, the variant's options are merged over the image's options, with any field not explicitly set on the variant inheriting the image-level value. SOCI conversion runs as a post-build step (for example, via `bakery soci convert` or the `--soci-mode` flag on `bakery ci publish`) and only acts on targets where `enabled` is `true`.
```
bschwedler:
```suggestion
SOCI options can be set in the `options` array of an [Image](#image) or an [ImageVariant](#imagevariant). When both are present, the variant's options are merged over the image's options, with any field not explicitly set on the variant inheriting the image-level value. SOCI conversion runs as a post-build step via `bakery soci convert` (standalone) or as part of `bakery ci publish` (config-driven), and only acts on targets where `enabled` is `true`.
```

posit-bakery/posit_bakery/cli/ci.py
```diff
@@ -238,51 +300,125 @@ def merge(
...
+            raise typer.Exit(code=1)
+        res = OrasIndexCreateWorkflow(
+            oras_bin=oras_bin,
+            image_target=t,
+            annotations=t.labels,
+        ).run(dry_run=dry_run)
+        if not res.success:
+            log.error(f"index-create failed for '{t}': {res.error}")
+            raise typer.Exit(code=1)
+        temp_refs[t.uid] = res.temp_ref  # type: ignore[assignment]
```
bschwedler: Remove the suppression — it only exists because of the `str | None` type on `OrasIndexCreateResult.temp_ref` (see suggestion on `oras.py:190`).

```suggestion
        temp_refs[t.uid] = res.temp_ref
```

bschwedler (changes requested):
This looks good to me.

My "Request changes" Is only to block merging until the the commit changing the branch is dropped from the PR.

bschwedler (approved):
Make it so.

PR #557, opened 2026-05-29
Remove ORAS manifest delete from merge workflow

ianpittwood:
## Summary

The `OrasManifestDelete` command frequently failed when cleaning up the temporary manifest index after a merge. This removes the command and the in-line cleanup step from `OrasMergeWorkflow`.

The temporary index is now left in place and cleaned up out-of-band by the `clean.yml` workflow (`bakery clean temp-registry`), which each product repo already runs on a schedule.

## Changes

- Removed the `OrasManifestDelete` class from `oras.py`.
- Removed "Step 3" (delete temporary index) from `OrasMergeWorkflow.run()`; the workflow now only creates the temporary index and copies it to destinations.
- Updated the workflow docstring/comments to note the temp index is cleaned up by `clean.yml`.
- Updated tests: dropped the `OrasManifestDelete` import, removed `TestOrasManifestDelete` and `test_manifest_delete_with_plain_http`, and adjusted `test_execute_success`'s call-count assertion from 4 → 3.

## Testing

- `uv run pytest test/plugins/builtin/oras/` → 43 passed
- `uv run ruff check` on both modified files → all checks passed

## Notes

With the in-line delete gone, temporary indexes now persist until the cleanup job runs (default: removes those older than 3 days) rather than being deleted immediately after each merge.

### DIFF
--- a/posit-bakery/posit_bakery/plugins/builtin/oras/oras.py
+++ b/posit-bakery/posit_bakery/plugins/builtin/oras/oras.py
@@ -142,24 +142,6 @@ class OrasCopy(OrasCommand):
         return cmd
 
 
-class OrasManifestDelete(OrasCommand):
-    """Delete a manifest from a registry.
-
-    This command deletes a manifest (image or index) from a registry.
-    """
-
-    reference: Annotated[str, Field(description="The manifest reference to delete.")]
-
-    @property
-    def command(self) -> list[str]:
-        """Build the oras manifest delete command."""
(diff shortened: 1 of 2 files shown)


bschwedler (approved):
Is there any reason we would want to keep this around as a standalone CLI command?

ianpittwood:
> Is there any reason we would want to keep this around as a standalone CLI command?

I can't think of any. It seems like GHCR is straight up incompatible with OCI deletion commands.

PR #565, opened 2026-06-04
Add --dev-spec for dispatch-pinned dev builds

### LINKED ISSUE #561: Add --dev-spec dispatch payload for pinned dev builds
When a product CI workflow dispatches a dev image build, bakery discovers the current dev version from the upstream CDN at matrix-generation time. The CDN upload is asynchronous, so during the propagation window bakery can resolve the previous build while the dispatch targets a newer one. The result is a silently wrong image: the download uses the stale artifact but the image is tagged with the dispatched version.

This issue tracks adding a typed JSON dispatch payload (`--dev-spec` / `BAKERY_DEV_SPEC`) that lets a dispatch pin the version exactly, bypassing CDN discovery:

```bash
bakery build --dev-spec '{"version": "2026.05.0-dev+185-gSHA", "channel": "daily"}'
```

## Design decisions

- Applied at `BakeryConfig.load()` before resolution, scoped to the dispatched channel, with an ambiguity guard (raises if multiple stream dev versions match and no channel disambiguates)
- PPM (template streams): URL rendered offline from the pinned version — no network fetch
- Connect / Workbench (manifest streams): fetch the manifest, assert version matches, raise loudly on mismatch (`DispatchVersionMismatchError`) instead of silently building stale
- Inert on non-dispatch runs (scheduled, PR, production, local) when `BAKERY_DEV_SPEC` is unset
- Payload is extensible: future fields (`sha`, `build_number`, `branch`) are additive-only, with zero workflow or `bakery.yaml` churn per addition

## Sequencing

- Phase 0: Remove dormant `env` dev-version source type (prerequisite cleanup)
- Phase 0.5a: Rename `stream` → `channel` in bakery internals and CLI (`--dev-stream` → `--dev-channel`), with backward-compat shim in shared workflows so product repos need no changes during implementation
- Phase 1: Implement `--dev-spec` / `DevBuildSpec` / shared workflow wiring
- Per-repo follow-up: Each product repo makes one PR switching to `channel` naming and wiring `dev-spec` on dispatch

Closes #527

## Follow-up PRs

- https://github.com/posit-dev/connect/pull/40410
- https://github.com/posit-dev/images-connect/pull/120
- https://github.com/posit-dev/images-package-manager/pull/105

bschwedler:
Upstream CI can now pin an exact dev version on workflow dispatch,
eliminating the CDN-propagation window where a re-triggered build picks
up a stale artifact instead of the one just built.

`--dev-spec` (or `BAKERY_DEV_SPEC`) accepts a JSON payload on `bakery
build` and `bakery ci matrix`. Template-URL products (PPM) render the
download URL offline from the pinned version. Manifest products (Connect,
Workbench) fetch and assert the manifest matches, raising loudly rather
than silently building stale.

Preparatory work: removes the dormant `env` dev-version source type, and
renames `stream` to `channel` throughout. `--dev-stream` and `stream:` in
bakery.yaml are retained as deprecated aliases.

### DIFF
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -9,7 +9,6 @@ import typer
 from posit_bakery.cli.common import with_verbosity_flags, with_temporary_storage, parse_dev_spec
 from posit_bakery.config import BakeryConfig
 from posit_bakery.config.config import BakeryConfigFilter, BakerySettings
-from posit_bakery.config.image.dev_version.spec import DevBuildSpec
 from posit_bakery.config.image.posit_product.const import ReleaseChannelEnum
 from posit_bakery.const import DevVersionInclusionEnum, MatrixVersionInclusionEnum
 from posit_bakery.error import BakeryBuildErrorGroup, BakeryToolRuntimeError
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -10,6 +10,7 @@ from posit_bakery.cli.common import with_verbosity_flags, with_temporary_storage
 from posit_bakery.config import BakeryConfig
 from posit_bakery.config.config import BakeryConfigFilter, BakerySettings
 from posit_bakery.config.image.posit_product.const import ReleaseChannelEnum
+from posit_bakery.config.image.posit_product.errors import DispatchVersionMismatchError
 from posit_bakery.const import DevVersionInclusionEnum, MatrixVersionInclusionEnum
 from posit_bakery.error import BakeryBuildErrorGroup, BakeryToolRuntimeError
 from posit_bakery.image import ImageBuildStrategy
@@ -253,7 +254,11 @@ def build(
         dev_spec=dev_spec,  # type: ignore[arg-type]  # typer requires str annotation; parse_dev_spec callback delivers DevBuildSpec at runtime
     )
 
-    config: BakeryConfig = BakeryConfig.from_context(context, settings)
+    try:
+        config: BakeryConfig = BakeryConfig.from_context(context, settings)
+    except DispatchVersionMismatchError as e:
+        stderr_console.print(f"❌ {e}", style="error")
+        raise typer.Exit(code=1)
 
     if plan:
         if strategy == ImageBuildStrategy.BUILD:
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -20,6 +20,17 @@ from posit_bakery.util import auto_path
 log = logging.getLogger(__name__)
 
 
+def _parse_dev_spec(ctx: typer.Context, param: typer.CallbackParam, value: str | None) -> DevBuildSpec | None:
+    if value is None:
+        return None
+    from pydantic import ValidationError
+
+    try:
+        return DevBuildSpec.model_validate_json(value)
+    except ValidationError as e:
+        raise typer.BadParameter(str(e), ctx=ctx, param=param) from e
+
+
 class RichHelpPanelEnum(str, Enum):
     """Enum for categorizing options into rich help panels."""
 
@@ -213,12 +224,13 @@ def build(
         ),
     ] = False,
     dev_spec: Annotated[
-        str | None,
+        Optional[str],
         typer.Option(
             "--dev-spec",
             envvar="BAKERY_DEV_SPEC",
             help='JSON spec for a dispatched dev build. Ex: \'{"version": "2026.05.0-dev+185-gSHA", "channel": "daily"}\'',
             rich_help_panel=RichHelpPanelEnum.FILTERS,
+            callback=_parse_dev_spec,
         ),
     ] = None,
 ) -> None:
@@ -235,12 +247,6 @@ def build(
         log.warning("--dev-stream is deprecated, use --dev-channel instead.")
         if dev_channel is None:
             dev_channel = dev_stream
-    parsed_dev_spec: DevBuildSpec | None = None
-    if dev_spec:
-        try:
-            parsed_dev_spec = DevBuildSpec.model_validate_json(dev_spec)
-        except Exception as e:
-            raise typer.BadParameter(f"--dev-spec is not valid JSON: {e}", param_hint="--dev-spec") from e
     settings = BakerySettings(
         filter=BakeryConfigFilter(
             image_name=image_name,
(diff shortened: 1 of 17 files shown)


posit-bakery/posit_bakery/cli/build.py
```diff
@@ -193,6 +204,15 @@ def build(
             rich_help_panel=RichHelpPanelEnum.FILTERS,
         ),
     ] = MatrixVersionInclusionEnum.EXCLUDE,
+    dev_spec: Annotated[
+        str | None,
```
ianpittwood, line 216: Let's try using a callback to validate and parse this value to the DevBuildSpec  object

posit-bakery/posit_bakery/config/image/posit_product/main.py
```diff
@@ -21,8 +21,16 @@
 from posit_bakery.util import cached_session
 
 
-class ReleaseStreamResult(BaseModel):
-    """Represents a resulting artifact found in a release stream. This provides an easy validation for data we get."""
+class DispatchVersionMismatchError(Exception):
```
ianpittwood: I'd say move this to an errors.py in this module or in the global errors.py

posit-bakery/posit_bakery/config/image/dev_version/spec.py
```diff
@@ -0,0 +1,20 @@
+from pydantic import BaseModel, ConfigDict, field_validator
+
+from posit_bakery.config.image.posit_product.const import ReleaseChannelEnum
+
+
+class DevBuildSpec(BaseModel):
```
ianpittwood, line 6: I would prefer if this also took a download url

posit-bakery/posit_bakery/config/image/version.py
```diff
@@ -134,22 +134,22 @@ class ImageVersion(BakeryPathMixin, BakeryYAMLModel):
...
         if not self.isDevelopmentVersion and dev_versions == DevVersionInclusionEnum.ONLY:
             return False, "not a development version (excluded by --dev-versions only)"
-        if dev_stream is not None and self.isDevelopmentVersion:
-            version_stream = self.metadata.get("release_stream")
-            if version_stream != dev_stream:
-                return False, (f"dev stream '{version_stream}' does not match --dev-stream '{dev_stream.value}'")
+        if dev_channel is not None and self.isDevelopmentVersion:
+            version_channel = self.metadata.get("release_channel")
+            if version_channel != dev_channel:
+                return False, (f"dev channel '{version_channel}' does not match --dev-channel '{dev_channel.value}'")
```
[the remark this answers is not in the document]
ianpittwood: We should fix this one

posit-bakery/posit_bakery/config/image/dev_version/channel.py
```diff
@@ -2,32 +2,54 @@
...
-    """Image development version sourced from a product stream."""
+    """Image development version sourced from a product release channel."""
 
     sourceType: Literal["stream"] = "stream"
-    product: Annotated[ProductEnum, Field(description="The ID of the product stream to use for this image version.")]
-    stream: Annotated[ReleaseStreamEnum, Field(description="The release stream to use for this image version.")]
+    product: Annotated[ProductEnum, Field(description="The ID of the product channel to use for this image version.")]
+    channel: Annotated[
+        ReleaseChannelEnum,
+        Field(description="The release channel to use for this image version (e.g. 'daily', 'preview')."),
```
ianpittwood, line 23: Add a backwards compatible alias here or I think everything breaks pointing to `main`

posit-bakery/posit_bakery/config/image/dev_version/channel.py
```diff
@@ -2,32 +2,54 @@
...
 from posit_bakery.config.image.version_os import ImageVersionOS
 
 log = logging.getLogger(__name__)
 
 
 class ImageDevelopmentVersionFromProductStream(BaseImageDevelopmentVersion):
-    """Image development version sourced from a product stream."""
+    """Image development version sourced from a product release channel."""
 
     sourceType: Literal["stream"] = "stream"
```
ianpittwood, line 19: should this also change or is that a separate PR?

posit-bakery/posit_bakery/cli/build.py
```diff
@@ -203,6 +223,16 @@ def build(
...
+    if dev_stream is not None:
+        log.warning("--dev-stream is deprecated, use --dev-channel instead.")
+        if dev_channel is None:
+            dev_channel = dev_stream
+    parsed_dev_spec: DevBuildSpec | None = None
+    if dev_spec:
+        try:
+            parsed_dev_spec = DevBuildSpec.model_validate_json(dev_spec)
+        except Exception as e:
+            raise typer.BadParameter(f"--dev-spec is not valid JSON: {e}", param_hint="--dev-spec") from e
```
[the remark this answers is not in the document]
ianpittwood: This can be handled as part of switching to a callback. I think we just need to catch whatever exception type Pydantic will throw

ianpittwood (changes requested):
I like the concept, but I have some other concerns

posit-bakery/posit_bakery/config/image/dev_version/spec.py
bschwedler: I'll implement this as a separate PR on top of this one.

posit-bakery/posit_bakery/config/image/dev_version/channel.py
bschwedler: Separated out into #572

posit-bakery/posit_bakery/config/image/posit_product/errors.py
```diff
@@ -0,0 +1,6 @@
+class DispatchVersionMismatchError(Exception):
```
ianpittwood, line 1: This is not properly caught when running `bakery build` against a version past what is available. It triggers a VERY lengthy stack trace which we should intercept and format better.

posit-bakery/posit_bakery/cli/common.py
```diff
@@ -7,19 +7,34 @@
...
 from posit_bakery.settings import SETTINGS
 
 log = logging.getLogger(__name__)
 
 
+def parse_dev_spec(ctx: typer.Context, param: typer.CallbackParam, value: str | None) -> DevBuildSpec | None:
+    if value is None:
+        return None
+    try:
+        data = json.loads(value)
```
ianpittwood: We could probably just use `model_validate_json`, but this might be intentional for cleaner error messaging?

ianpittwood (approved):
I think this looks good, but I would really like to see the absurd stacktrace on `DispatchVersionMismatchError` fixed.

posit-bakery/posit_bakery/cli/common.py
bschwedler: I added a step in `model_validate_json` to raise the exception if JSON parsing fails.

posit-bakery/posit_bakery/config/image/posit_product/errors.py
bschwedler: This has been addressed.

PR #566, opened 2026-06-04
feat(ci): per-image merge jobs in native build workflow

ianpittwood:
## Summary

- The `merge` job in `bakery-build-native.yml` previously ran as a single job for all images built by the workflow. It now fans out as a **matrix keyed by image name** — one merge job per distinct image (e.g. separate `workbench` and `workbench-session-init` merges).
- Added an `--image-name` regex filter to `bakery ci merge` (mirroring `bakery ci matrix` / `bakery build`), so each per-image merge job scopes itself with `--image-name "^${IMAGE_NAME}$"`. The anchored regex keeps prefix-colliding names isolated (`workbench` won't match `workbench-session-init`).
- The `matrix` job now also emits an `image-matrix` output (`jq '[.[].image] | unique'`) that the merge matrix iterates over.

This is about parallelism, isolation, and clearer per-image logs — `bakery ci merge` already only acted on targets whose metadata was present, so correctness is unchanged.

## Why it's safe for caller repos

No `workflow_call` inputs changed. The additions are internal to the shared workflow (a new `matrix` output, a `strategy.matrix` on `merge`, and a `needs` addition). Callers pinning `@main` get the matching `bakery` via `setup-bakery@main`, so the CLI flag and workflow stay in sync. No changes required in `images-connect`, `images-workbench`, or `images-package-manager`.

## Test Plan

- [x] `uv run pytest test/cli/test_ci.py` — 27 passed, including three new merge scenarios:
  - matching image name → full merge
  - non-matching anchored name → 0 targets (no-op-filter guard)
  - **prefix-colliding** fixture (`test-alpha` vs `test-alpha-extra`) → only `test-alpha` merged (selectivity / anchoring guard)
- [x] `actionlint` passes on `bakery-build-native.yml` (via pre-commit)
- [ ] Confirm a real native build run produces one merge job per image and pushes correct multi-platform manifests

### DIFF
--- a/.github/workflows/bakery-build-native.yml
+++ b/.github/workflows/bakery-build-native.yml
@@ -104,6 +104,7 @@ jobs:
       contents: read
     outputs:
       platform-matrix: ${{ steps.images-by-platform.outputs.platform_matrix }}
+      image-matrix: ${{ steps.images-by-platform.outputs.image_matrix }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


bschwedler:
Does this still push the images to Docker Hub in reverse order so the latest pushed tag is correct?

ianpittwood:
> Does this still push the images to Docker Hub in reverse order so the latest pushed tag is correct?

Pushes are still sorted by version per image so it shouldn't have an impact on ordering.

bschwedler (approved):

PR #567, opened 2026-06-04
Add dependency-backed dev version and Positron daily channel

bschwedler:
Positron's CDN URL is arch-parameterized, which doesn't fit the existing product channel model. Routing through the dependency constraint system reuses the arch-handling logic already in `PositronDependency` rather than duplicating it.

Stacked on:
- https://github.com/posit-dev/images-shared/pull/565


ianpittwood (changes requested):
This creates a bit of weirdness where Positron is considered as both a dependency and a product. We'll need to make it live in one of those worlds. I would vote for keeping it a dependency and finding a way to implement dev version or prerelease behaviors with dependency matrix builds.

I would still like a `--latest` flag for building matrix builds to single out the latest image only. Maybe a similar concept could be implemented for `--pre-release` to build the an image with dependency pre-releases if defined/available?

ianpittwood:
Another idea, we could move it fully into the product world and make a section for `matrix.productConstraints` that would work through the product resolver workflow.

ianpittwood (changes requested):
I think the mechanism makes more sense now, but we're missing docs updates. I would also like to possibly see a test suite or a more rigidly defined example of the configuration in action.

bschwedler:
> I think the mechanism makes more sense now, but we're missing docs updates. I would also like to possibly see a test suite or a more rigidly defined example of the configuration in action.

I've updated the documentation and extended the test suite.

ianpittwood:
> > I think the mechanism makes more sense now, but we're missing docs updates. I would also like to possibly see a test suite or a more rigidly defined example of the configuration in action.
> 
> I've updated the documentation and extended the test suite.

Uhhhh I don't see any commits since my last review? Did they not push or is GitHub having a goof?

bschwedler:
> Uhhhh I don't see any commits since my last review? Did they not push or is GitHub having a goof?

I forgot to push 🤦🏻

ianpittwood (approved):

PR #569, opened 2026-06-05
Add --latest flag to bakery build

ianpittwood:
## Summary
- Adds a `--latest` filter to `bakery build` that builds only the latest version of each image.
- Standard versions: the version marked `latest: true`. Matrix versions: the combination identified by the existing matrix latest-determination logic (so `--latest --matrix-versions include` builds only the latest combination).
- Development versions are always ignored by `--latest`; combining with `--dev-versions only`/`include` logs a warning and proceeds (warn-and-continue, no hard error).

## Implementation
- `BakerySettings` gains a `latest: bool = False` field.
- `generate_image_targets` skips any version that is not `version.latest` or is a development version when `--latest` is set, mirroring the existing skip/log idiom (warn if it collides with an explicit `--image-version` filter, else debug).
- `bakery build` gains a `--latest/--no-latest` option in the Filters help panel and emits the dev-versions warning.

## Test Plan
- [x] `uv run pytest test/config/test_config.py -k latest` (standard, matrix, dev exclusion, `--image-version` collision warning, settings default)
- [x] `uv run pytest test/cli/test_build.py -k TestBuildLatestFlag` (settings pass-through, warning emission/suppression)
- [x] `just test` — 1578 passed
- [x] `bakery build --help` shows `--latest / --no-latest` under Filters

### DIFF
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -1,4 +1,3 @@
-import logging
 from enum import Enum
 from pathlib import Path
 from typing import Annotated, Optional
@@ -6,7 +5,11 @@ from typing import Annotated, Optional
 import python_on_whales
 import typer
 
-from posit_bakery.cli.common import with_verbosity_flags, with_temporary_storage
+from posit_bakery.cli.common import (
+    with_verbosity_flags,
+    with_temporary_storage,
+    warn_if_latest_ignores_dev_versions,
+)
 from posit_bakery.config import BakeryConfig
 from posit_bakery.config.config import BakeryConfigFilter, BakerySettings
 from posit_bakery.config.image.posit_product.const import ReleaseStreamEnum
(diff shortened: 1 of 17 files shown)


posit-bakery/posit_bakery/cli/build.py
```diff
@@ -193,6 +193,14 @@ def build(
             rich_help_panel=RichHelpPanelEnum.FILTERS,
         ),
     ] = MatrixVersionInclusionEnum.EXCLUDE,
+    latest: Annotated[
+        Optional[bool],
+        typer.Option(
+            "--latest/--no-latest",
```
bschwedler: What does the `--no-latest` do in this case? Just build all of them?

bschwedler (approved):

posit-bakery/posit_bakery/cli/build.py
ianpittwood: Sorry I forgot to tell Claude to do this as a draft! I will get rid of that, it's unnecessary.

PR #573, opened 2026-06-08
fix: remove stale hook config definition

ianpittwood:


ianpittwood:
Should've done this in #570 but I forgot

bschwedler (approved):

Issue #575, opened 2026-06-08
2026.b: 🏗️ RHEL 10 (UBI) image support


bschwedler:
Build and validate RHEL container images.

## Background

First-party Posit product images ship on Ubuntu only today. RHEL/UBI
(`registry.access.redhat.com/ubiN/ubi`) support has never been attempted
for Connect or Workbench; Package Manager has one in-flight attempt
(images-package-manager#103, not yet merged).

Depends on:
- https://github.com/posit-dev/platform-team/issues/443

## Goals

- Standard and Minimal builds of package-manager, connect, and workbench
  on RHEL 10 (UBI), amd64 + arm64
- RHEL 10 support for the connect-content/connect-content-init and
  workbench-session/workbench-session-init matrix and init images
- The images EKS smoke-test pipeline (platform-team#443) validates
  RHEL10/UBI image variants the same way it validates Ubuntu images

## External Dependencies and Blockers

- rstudio/r-builds#329 (merged) already fixes the `flexiblas-devel` hard
  dependency that blocks R on UBI 9/10 — consuming repos just need R
  4.6.1+ to pick it up.
- platform-infra#293 is tracked under the test-automation epic
  (platform-team#443), not natively under this epic.

## Notes

Phase 1 (package-manager) in progress: images-package-manager#103,
tracked via images-package-manager#130.

bschwedler:
R installation requires the `flexiblas-devel` package, which is not available in the CodeReady Builder (`cbr`) repo for the Universal Baseline Image
- https://github.com/rstudio/r-builds/issues/212

PR #577, opened 2026-06-08
feat(r.j2): add manylinux R tarball install macros for self-contained R

### LINKED ISSUE #747: r.j2: add manylinux R tarball install macros
Add `install_manylinux()` / `run_install_manylinux()` to `r.j2`, so R can be
installed from a self-contained `manylinux_2_34` tarball instead of the
deb/RPM package.

## Why

The deb/RPM install dynamically links against system libraries (`libcairo2`,
`libcurl4t64`, `libicu-dev`, `libopenblas-dev`, `libpango-1.0-0`,
`libtcl8.6`, `libtk8.6`, `libx11-6`, etc.) that live in the installing OS's
own package tree. That's fine for a normal product image that ships its full
OS alongside R, but breaks for posit-dev/images-volumes, whose images are
built `FROM scratch` and only carry `/opt/R` — none of those system
libraries make the cut, so R fails at runtime once mounted into a container
that doesn't happen to have Ubuntu's exact package set installed. Verified
directly: copying only `/opt/R` into an unrelated `debian:12-slim` container
and running `R` fails with `error while loading shared libraries:
libblas.so.3: cannot open shared object file`.

The `manylinux_2_34` tarball at `cdn.posit.co/r/manylinux_2_34/` bundles its
own non-glibc runtime dependencies (RUNPATH `$ORIGIN/../../lib/.libs`) and
only requires glibc >= 2.34 from the host, so it's safe to copy alone into a
`scratch` image.

## Notes

Originally written as a workaround for the RHEL 10 (UBI) `flexiblas-devel`
blocker (posit-dev/images-shared#575), but that's fixed upstream
(rstudio/r-builds#329, merged) and doesn't need this. The macros are being
kept for the unrelated `images-volumes` use case instead.

## Related

- https://github.com/posit-dev/images-shared/pull/577

bschwedler:
Adds `install_manylinux()` and `run_install_manylinux()` alongside the existing RPM/deb path. Downloads pre-built R tarballs from `cdn.posit.co/r/manylinux_2_34/` and extracts them to `/opt/R/{version}/`, bypassing package manager dependency resolution. Arch is detected at runtime; the same Containerfile covers amd64 and arm64.

Useful for posit-dev/images-volumes, whose images are built `FROM scratch` and only carry `/opt/<tool>` — the manylinux tarball is self-contained (bundles its own non-glibc runtime deps, only requires glibc >= 2.34), unlike the deb/RPM path which dynamically links against system libraries that don't survive the copy into `scratch`.

Closes https://github.com/posit-dev/images-shared/issues/747

### DIFF
--- a/posit-bakery/posit_bakery/config/templating/macros/r.j2
+++ b/posit-bakery/posit_bakery/config/templating/macros/r.j2
@@ -26,31 +26,35 @@ RUN_UNATTENDED=1 R_VERSION={{ version | trim }} bash -c "$(curl -fsSL https://rs
 find . -type f -name '[rR]-{{ version | trim }}.*\.(deb|rpm)' -delete
 {%- endmacro %}
 
-{# Downloads and extracts a manylinux_2_34 R tarball from cdn.posit.co/r directly to
-/opt/R/{version}/. Requires glibc >= 2.34 (RHEL 10, Ubuntu 22.04+, Debian 12+).
-Use in place of run_install when the platform cannot satisfy the RPM/deb dependencies.
+{# Downloads and extracts a manylinux R tarball from cdn.posit.co/r directly to
+/opt/R/{version}/. Use in place of run_install when the platform cannot satisfy
+the RPM/deb dependencies.
 
 :param version: The R version to install, e.g. "4.4.3"
+:param glibc_version: The manylinux glibc version string, e.g. "2_34" (default).
+    Requires the host glibc to be >= the chosen version.
 #}
-{% macro install_manylinux(version) -%}
+{% macro install_manylinux(version, glibc_version="2_34") -%}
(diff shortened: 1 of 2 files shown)


bschwedler:
Re-scoping this off the RHEL 10 (UBI) epic (posit-dev/images-shared#575) — that blocker is fixed upstream (rstudio/r-builds#329, merged) and doesn't need this workaround; posit-dev/images-package-manager#104 has been closed accordingly.

These macros are still useful on their own merits: the manylinux tarball is self-contained (bundles its non-glibc runtime deps, only requires glibc >= 2.34), which matters for posit-dev/images-volumes' scratch-based images that get copied alone into arbitrary containers. Tracking under a new epic instead.

PR #578, opened 2026-06-09
Add scratch OS support

### LINKED ISSUE #386: Support OS-less images in Bakery
## Problem

Bakery requires every image version to have at least one OS entry with non-empty `extension` and `tagDisplayName` fields. Both fields enforce regex patterns that require at least one character (`^[a-zA-Z0-9_-]+$` and `^[a-zA-Z0-9_.-]+$`), making empty strings invalid.

This prevents building images that don't need an OS, such as:
- Init containers based on `scratch` that only copy static files
- Utility images with statically-linked Go binaries
- Any image where the final stage is `FROM scratch`

## Context

The `workbench-session-init` image was recently refactored to use `FROM scratch` since it only contains pre-built static binaries from the `rsp-session-multi` tarball. The builder stage uses Ubuntu to download and extract the tarball, but the final image has no OS. Despite this, Bakery still requires an OS entry (`Ubuntu 24.04`), resulting in misleading tags like `workbench-session-init:2026.01.1-ubuntu-24.04` for an image that contains no Ubuntu.

Relevant code: `posit_bakery/config/shared.py` lines 14-34 define `ExtensionField` and `TagDisplayNameField` with patterns that reject empty strings.

## Proposed solution

Allow OS `extension` and `tagDisplayName` to be empty strings (or omitted entirely) so that:
- The Containerfile has no OS suffix (e.g., just `Containerfile.jinja2`)
- Tags omit the OS component (e.g., `workbench-session-init:2026.01.1` with no `-ubuntu-24.04` suffix)

This could be implemented by:
1. Making the regex patterns accept empty strings: `^[a-zA-Z0-9_.-]*$`
2. Or adding a new OS-less image mode that skips OS-related naming entirely

## References

- PR: https://github.com/posit-dev/images-workbench/pull/41

bschwedler:
Enables images that don't vary by OS — no OS suffix in tags or Containerfile names. The `artifactOs` field handles the dev-channel case where the channel API has no scratch-native artifact entries.


.claude/settings.json
```diff
@@ -2,7 +2,6 @@
   "enabledPlugins": {
     "pyright-lsp@claude-plugins-official": true
   },
-  "hooks": {},
```
ianpittwood, line 5: Automatic change by Claude or something?

.github/workflows/bakery-build-pr.yml
ianpittwood, line 1: This seems unrelated...

posit-bakery/posit_bakery/cli/ci.py
ianpittwood, line 1: Yep I think this branch must have been polluted...

posit-bakery/posit_bakery/config/image/version_os.py
```diff
@@ -33,16 +66,30 @@ class ImageVersionOS(BakeryYAMLModel):
...
     extension: Annotated[
-        ExtensionField,
+        str,
         Field(
+            default_factory=lambda data: (
+                ""
+                if data.get("name", "").lower().strip() in _OSLESS_NAMES
+                else re.sub(r"[^a-zA-Z0-9_-]", "", data.get("name", "").lower())
+            ),
+            pattern=r"^[a-zA-Z0-9_-]*$",
```
ianpittwood: I think at least part of this is still shared. Extract to constant?

posit-bakery/posit_bakery/config/image/version_os.py
```diff
@@ -33,16 +66,30 @@ class ImageVersionOS(BakeryYAMLModel):
...
     tagDisplayName: Annotated[
-        TagDisplayNameField,
+        str,
         Field(
+            default_factory=lambda data: (
+                ""
+                if data.get("name", "").lower().strip() in _OSLESS_NAMES
+                else re.sub(REGEX_IMAGE_TAG_SUFFIX_ALLOWED_CHARACTERS_PATTERN, "-", data.get("name", "").lower())
+            ),
+            pattern=r"^[a-zA-Z0-9_.-]*$",
```
ianpittwood: Same, constant?

ianpittwood (changes requested):
A lot of this is making sense, but the branch seems like it was polluted with another change

.claude/settings.json
bschwedler: Yep. Two sessions not using separate worktrees. 🤖

PR #579, opened 2026-06-09
Mask credentials in plugin tool logging

bschwedler:
Secret values (wizcli client secret, GH_TOKEN) were appearing unredacted in process listings, log output, and error metadata.

### DIFF
--- a/posit-bakery/posit_bakery/util.py
+++ b/posit-bakery/posit_bakery/util.py
@@ -9,6 +9,46 @@ from requests_cache import CachedSession
 
 from posit_bakery.error import BakeryToolNotFoundError
 
+
+class SensitiveArg:
+    """Wraps a CLI argument value that must not appear in logs or error output.
+
+    Use for credentials passed as command-line flags (e.g. ``--client-secret VALUE``).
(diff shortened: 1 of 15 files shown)


posit-bakery/posit_bakery/util.py
```diff
@@ -9,6 +9,46 @@
 
 from posit_bakery.error import BakeryToolNotFoundError
 
+
+class SensitiveArg:
```
ianpittwood, line 13: This feels excessive for something we only use once. I'm wondering if we could do per-plugin inline replacements or a smaller wrapper function that either replaces the value with `*`s or an output like `(hidden, X characters)`.

ianpittwood (changes requested):
I would prefer to make a `display_command` function for each plugin's command model or create a wider spanning `ShellCommand` class with methods for `to_shell()` and `to_display()`.

I do think we should globally mask `GH_TOKEN` and `GITHUB_TOKEN` by default regardless of whether its used or defined for an image.

The current implementation seems to accomplish what we're going for. I only think we could do it a bit cleaner.

PR #581, opened 2026-06-09
Notify on merge queue failures

bschwedler:
The Slack notification action previously skipped silently when no prior
run history was found. Merge queue runs on ephemeral refs never appear
in a main-branch run query, so that guard was swallowing all merge queue
failures with no notification.

Now: no history + passing → still skip. No history + failing → notify.

### DIFF
--- a/.github/actions/slack-build-notify/action.yml
+++ b/.github/actions/slack-build-notify/action.yml
@@ -54,32 +54,36 @@ runs:
               .filter(r => r.id !== context.runId && r.conclusion !== 'cancelled')
               .sort((a, b) => new Date(b.created_at) - new Date(a.created_at));
 
-            if (previousRuns.length === 0) {
-              core.info('No previous completed runs found, skipping notification');
-              return;
-            }
-
-            const previousRun = previousRuns[0];
             const currentFailed = currentResult !== 'success';
 
-            // Compare against the previous run's CI job conclusion (not the workflow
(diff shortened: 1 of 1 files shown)


.github/workflows/bakery-build-pr.yml
```diff
@@ -102,9 +107,12 @@ jobs:
           IS_FORK: ${{ needs.detect.outputs.is-fork }}
           DEV_VERSIONS: ${{ inputs.dev-versions }}
           MATRIX_VERSIONS: ${{ inputs.matrix-versions }}
+          LATEST_ONLY: ${{ inputs.latest-only }}
           BAKERY_CONTEXT: ${{ inputs.context }}
         run: |
-          FULL_MATRIX=$(bakery ci matrix --quiet --dev-versions "$DEV_VERSIONS" --matrix-versions "$MATRIX_VERSIONS" --context "$BAKERY_CONTEXT" | jq --compact-output .)
+          LATEST_FLAGS=()
+          if [ "$LATEST_ONLY" = "true" ]; then LATEST_FLAGS=(--latest); fi
```
bschwedler: I want to be able to run only the latest version of every image for each PR.

This allows us to enable merge queues in all the repos. Then when ready to merge it runs the full suite without passing `--latest`.

bschwedler: I am moving `latest` out into a separate PR.

ianpittwood (approved):
LGTM

PR #585, opened 2026-06-09
Add bakery Claude skill with inspect_ai evals

bschwedler:
Adds a Claude Code skill for the bakery CLI and an inspect_ai eval harness to validate it.

## What's included

**`.claude/skills/bakery/SKILL.md`** — A skill that encodes 13 critical invariants for working in Posit container image repos: never editing rendered files, always using `uv run`, reading `bakery.yaml` before constructing filter flags, forwarding flags consistently to `ci merge`, and more. Available in all sibling repos that have `images-shared` as an additional working directory.

**`.claude/skills/bakery/evals/`** — An inspect_ai eval harness with 13 dataset entries covering the skill's key invariants. Uses a behavior scorer that checks expected behaviors and forbidden behaviors against a grader model's assessment.

**`.github/workflows/bakery-skill-eval.yml`** — CI workflow that runs the evals when skill files change, uploads results as artifacts. Path-filtered to `.claude/skills/bakery/**` so it only fires on skill changes. Not wired into the required CI gate — evals are a signal, not a hard block.


ianpittwood (approved):
Seems like a great idea! Hopefully this helps fix some of the woes with multi-repo edits.

PR #588, opened 2026-06-09
feat: parallel dgoss execution via shared ParallelShellExecutor

ianpittwood:
## Summary

Adds bounded-concurrency parallel execution of `dgoss` test commands, built on a new **tool-agnostic** `posit_bakery/parallel/` module so other plugins (hadolint, wizcli) and future sequential builds can adopt the same pattern.

- **`posit_bakery/parallel/`** — `ShellTask`, `ShellResult`, `resolve_max_workers`, and `ParallelShellExecutor` (a `ThreadPoolExecutor`-backed runner). Worker threads only run their subprocess and return captured bytes; the main thread drains completions, drives an optional Rich live status table, and invokes a per-result callback — so consumers mutate shared state without locks. Results are returned in input order; spawn failures are captured (never crash the run).
- **Concurrency control** — `--jobs/-j` CLI flag on `bakery dgoss run` → `BAKERY_MAX_CONCURRENCY` env / `SETTINGS.max_concurrency` → default of `4` (modest, since each task is a Docker container). Malformed env values fall back to the default with a warning instead of crashing the CLI.
- **Streaming/logging** — the live `Progress` table shares the same `stderr_console` as the logger, so per-task detail (flushed on completion, on the main thread) renders cleanly above the live region with no interleaving. The table auto-disables on non-TTY/CI, `--quiet`, or single-task runs, falling back to plain logging.
- **dgoss** — `DGossSuite.run()` now dispatches through the executor while preserving its exact `(GossJsonReportCollection, errors)` return contract; per-result parse/persist/classify logic is unchanged.

Design spec and implementation plan live under `docs/superpowers/` (gitignored, not committed).

## Test Plan

- [x] `just test` — 1639 passed
- [x] New unit tests: executor (ordering, bounded concurrency, spawn-failure capture, env/cwd, main-thread callback, live-display policy), `resolve_max_workers`, settings env parsing (incl. malformed/empty fallback)
- [x] dgoss suite tests: mocked parallel happy-path, `jobs`→`max_workers` wiring, spawn-failure → `BakeryDGossError`
- [x] CLI test: `dgoss run --jobs 2` forwards `jobs=2` through to the suite
- [ ] Manual smoke (requires Docker + built images): `bakery dgoss run --jobs 2 -v` shows the live status table + per-target detail, then the Goss summary

### DIFF
--- a/posit-bakery/posit_bakery/const.py
+++ b/posit-bakery/posit_bakery/const.py
@@ -35,3 +35,5 @@ POSIT_LABEL_PREFIX = "co.posit.image"
 OCI_LABEL_PREFIX = "org.opencontainers.image"
 
 JINJA2_TEMPLATE_EXTENSIONS = {".jinja2", ".j2", ".jinja"}
+
+DEFAULT_MAX_CONCURRENCY = 4
--- a/posit-bakery/posit_bakery/parallel/__init__.py
+++ b/posit-bakery/posit_bakery/parallel/__init__.py
@@ -1,0 +1,13 @@
+from posit_bakery.parallel.executor import (
+    ParallelShellExecutor,
+    ShellResult,
+    ShellTask,
+    resolve_max_workers,
+)
+
+__all__ = [
+    "ParallelShellExecutor",
+    "ShellResult",
+    "ShellTask",
+    "resolve_max_workers",
+]
--- a/posit-bakery/posit_bakery/parallel/executor.py
+++ b/posit-bakery/posit_bakery/parallel/executor.py
@@ -1,11 +1,8 @@
 from __future__ import annotations
 
-import subprocess
-import time
-from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable
+from typing import Any
 
 from posit_bakery.settings import SETTINGS
 
--- a/posit-bakery/posit_bakery/parallel/executor.py
+++ b/posit-bakery/posit_bakery/parallel/executor.py
@@ -104,6 +104,7 @@ class ParallelShellExecutor:
         self._use_live = use_live
         self._active: set[subprocess.Popen] = set()
         self._lock = threading.Lock()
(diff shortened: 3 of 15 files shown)


ianpittwood:
### Follow-up: per-task timeout + responsive interruption

Added on top of the initial parallel-dgoss work (4 commits, `b7b5c1c..bf28a59`):

**Responsive interruption** — `ParallelShellExecutor` now launches each child with `start_new_session=True` and tracks live processes in a lock-guarded registry. On `KeyboardInterrupt`/`BaseException` it cancels queued tasks and signals the whole **process group** (SIGTERM → 10s grace → SIGKILL) of every in-flight child, then re-raises. This fixes the prior behavior where Ctrl-C drained the full queue and could leave `docker`/dgoss containers running. Group-signaling (vs. signaling only the dgoss wrapper) is what prevents orphaned `docker` grandchildren; SIGTERM-first lets dgoss's EXIT trap remove its container. A registration-race window (a child spawned exactly as shutdown begins) is closed deterministically via a shutdown flag checked under the same lock.

**Per-task timeout** — `ShellTask.timeout` enforced via `Popen.communicate(timeout=…)`; on expiry the child group is stopped and the result flagged `timed_out`. Exposed on dgoss as **`GossOptions.timeout` (default 900s = 15 min)**, resolved through `DGossCommand` → `ShellTask` (0 disables). A timed-out target is reported as a run-failing `BakeryDGossError`.

**Tests** — real-subprocess executor tests for timeout, SIGINT cancellation, whole-process-group termination (heartbeat-divergence), and the shutdown-race path; dgoss suite tests for timeout→error and timeout plumbing. Note: this also repaired two dgoss suite tests that the `subprocess.run`→`Popen` switch had silently broken (they now mock `ParallelShellExecutor.run`). `just test`: 1649 passed.

ianpittwood:
Error appears related to Python release desync with `uv`. It should succeed once the available versions are back in sync.

bschwedler (approved):
💜

PR #590, opened 2026-06-10
Add release_branch field to channel dev version

bschwedler:
Adds a `release_branch` field to `ImageDevelopmentVersionFromProductChannel` so
the Workbench dailies branch can be configured directly in `bakery.yaml` instead
of relying on the `BAKERY_WORKBENCH_RELEASE_BRANCH` environment variable.

The field threads through all `get_product_artifact_by_channel()` call sites.
When a version is supplied via `--dev-spec`, the YYYY.MM segment is extracted
automatically and used as the release branch. When only `release_branch` is set
in the dev spec, it is assigned directly.

Also adds a `release-branch` input to `bakery-build-native.yml` so product repo
workflows can override the branch at dispatch time.

### DIFF
--- a/posit-bakery/posit_bakery/config/image/posit_product/main.py
+++ b/posit-bakery/posit_bakery/config/image/posit_product/main.py
@@ -17,7 +17,7 @@ from posit_bakery.config.image.posit_product.const import (
     CONNECT_DAILY_URL,
     DOWNLOADS_JSON_URL,
 )
-from posit_bakery.config.image.posit_product.errors import DispatchVersionMismatchError
+from posit_bakery.config.image.posit_product.errors import ArtifactNotAvailableError, VersionSubstitutionError
 from posit_bakery.config.shared import OSFamilyEnum
 from posit_bakery.util import cached_session
 
@@ -27,6 +27,7 @@ class ReleaseChannelResult(BaseModel):
 
     version: Annotated[str, Field(pattern=CALVER_REGEX_PATTERN)]
     download_url: HttpUrl
+    channel_latest: bool = True
 
     @computed_field
     @property
@@ -53,6 +54,7 @@ class ReleaseChannelPath:
         channel_url = self.channel_url.format_map(metadata)
 
         if version_override is not None and self.version_templatable:
+            # PPM: build artifact URL from template, then probe it and check channel head.
             result: dict = {"version": version_override}
             url_safe_result = {f"url_safe_{k}": quote(str(v), safe="") for k, v in result.items()}
             for key, resolver in self.resolver_map.items():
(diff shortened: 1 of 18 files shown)


zachhannum (approved):
LGTM!

posit-bakery/posit_bakery/config/image/posit_product/main.py
```diff
@@ -295,5 +298,6 @@ def get_product_artifact_by_channel(
         raise ValueError(f"Channel {channel} is not supported for product {product}.")
```
ianpittwood, line 355: We should add a warning about `release_branch` being ineffectual if `product != ProductEnum.WORKBENCH`. Not sure the best place to put the check. I guess `and release_branch != "latest"` could be another part of the `if` statement to make it work here?

ianpittwood (approved):
Unrelated to this change specifically, but the dev spec appears to not work against any daily build that isn't exactly what is current. This makes it effectively the same as the previous daily build implementation, doesn't it?

```
❌ Dispatched version '2026.06.0-daily+173.pro12' does not match manifest version '2026.06.0-daily+183.pro2' at 'https://dailies.rstudio.com/rstudio/2026.06/index.json'. The upstream manifest has not propagated the dispatched build yet.
```

I did ensure using something like `"release_branch": "golden-wattle"` works as expected.

bschwedler:
> Unrelated to this change specifically, but the dev spec appears to not work against any daily build that isn't exactly what is current. This makes it effectively the same as the previous daily build implementation, doesn't it?
> 
> ```
> ❌ Dispatched version '2026.06.0-daily+173.pro12' does not match manifest version '2026.06.0-daily+183.pro2' at 'https://dailies.rstudio.com/rstudio/2026.06/index.json'. The upstream manifest has not propagated the dispatched build yet.
> ```

I intentionally put this safety net #565. When implementing, I felt that failing louder early would be good.

The alternatives were:
- Replacing the version string in the URL
- Having it 404 later in the build process

Do you have a preference for how this would look?

ianpittwood:
> > Unrelated to this change specifically, but the dev spec appears to not work against any daily build that isn't exactly what is current. This makes it effectively the same as the previous daily build implementation, doesn't it?
> > ```
> > ❌ Dispatched version '2026.06.0-daily+173.pro12' does not match manifest version '2026.06.0-daily+183.pro2' at 'https://dailies.rstudio.com/rstudio/2026.06/index.json'. The upstream manifest has not propagated the dispatched build yet.
> > ```
> 
> I intentionally put this safety net #565. When implementing, I felt that failing louder early would be good.
> 
> The alternatives were:
> 
>     * Replacing the version string in the URL
> 
>     * Having it 404 later in the build process
> 
> 
> Do you have a preference for how this would look?

It makes sense to me to check if what's posted on daily is behind what the user is attempting to build, but if the user is attempting to build something older, I don't think it should fail. Does that make sense? 

To me if you can't build an arbitrary version, then we only added a complex entry to the original dev version retrieval logic. If there needs to be a pre-check, I would suggest making a request against the resolved URL. If it returns 4XX/5XX, then we error.

ssinnott (approved):

ssinnott (approved):
Nice!

PR #597, opened 2026-06-11
Run uv self update before Python install in build stage

bschwedler:
Adds `uv self update` before `uv python install` in the Python build-stage macro so the build always uses the latest uv when installing Python, rather than whatever version ships in the base image.

Updates tests and rendered test fixtures to match. Also gitignores `refs/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -192,6 +192,7 @@ out/
 .claude/settings.local.json
 .claude/worktrees/
 .claude/plans/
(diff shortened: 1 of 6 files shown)


bschwedler:
Closing — uv self update doesn't work in the astral uv base image since the binary was installed by the image, not the standalone installer. Will revisit with a different approach.

PR #600, opened 2026-06-11
Fix dependency version parsing

bschwedler:
Fixes the warning we see:

```
           WARNING  Unparseable version string:             parsed_version.py:58
                    'R4.6.0-python3.14.6'
```

- **Recognize dep-prefixed versions in ParsedVersion**
- **Parse dep-prefixed versions into separate components**


ianpittwood:
This was included by #601

PR #612, opened 2026-06-15
fix: stop silent dgoss/build workflow failures on version filters

### LINKED ISSUE #611: `bakery dgoss run` silently fails in all workflows
It seems like something related to version parsing:

```
           WARNING  Unparseable version string:             parsed_version.py:84
                    'R4\\.6\\.0\\-python3\\.14\\.6'                             
```

ianpittwood:
Fixes #611

## Summary

CI workflows were silently passing during `bakery dgoss run` while testing nothing, accompanied by warnings like:

```
WARNING  Unparseable version string: '2026\.01\.2\+418\.pro1'   parsed_version.py:84
```

This PR fixes the root cause and adds a guard so the failure can never be silent again.

### Root cause

Commit `50575feb` migrated version filtering from `re.search()` (which needed `re.escape()` on the filter) to segment-aware `version_matches()`, and updated the CLI call sites — but missed the `dgoss`, `hadolint`, and `wizcli` plugins. They still called `re.escape(image_version)`, mangling `2026.01.2+418.pro1` into `2026\.01\.2\+418\.pro1`. That failed to parse and failed the fallback comparison → zero targets matched → `dgoss run` tested nothing and exited 0.

### Changes

- **Remove `re.escape()`** from the `image_version` filter in the dgoss, hadolint, and wizcli plugins (and the now-unused `import re`), matching how the CLI commands already pass it through verbatim.
- **Add a shared `exit_if_no_targets()` guard** (`cli/common.py`) that aborts with a non-zero exit code, echoing the active filters, when filters resolve to zero image targets. Wired into `build`, `dgoss run` (+ deprecated `run dgoss`), `hadolint run`, and `wizcli scan`.

### Testing

- New regression test asserting `--image-version` reaches the filter unescaped.
- New zero-match guard tests for every affected command (exit 1, helpful message, execute/build never called).
- All affected suites pass with `-n auto`; ruff clean; verified end-to-end against the `basic` resource context.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,6 +176,7 @@ jobs:
       packages: write
     uses: "./.github/workflows/bakery-build-pr.yml"
     with:
+      version: ${{ github.head_ref || github.ref_name }}
       context: "./posit-bakery/test/resources/multiplatform/"
       dev-versions: include
 
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -6,7 +6,7 @@ from typing import Annotated, Optional
 import python_on_whales
 import typer
 
-from posit_bakery.cli.common import with_verbosity_flags, with_temporary_storage, parse_dev_spec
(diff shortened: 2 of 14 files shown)


bschwedler:
Thanks for fixing this.

PR #622, opened 2026-06-17
Seed CDN cache from matrix job into build jobs

### LINKED ISSUE #304: Dev version builds occasionally fail in native builder workflows

bschwedler:
The matrix step resolves dev versions from the CDN and outputs them as a build matrix. Each subsequent job runs on a fresh runner and re-queries the CDN independently. If the daily CDN updates between the matrix step and a build or merge step, the resolved version changes and the `--image-version` filter no longer matches any target, producing a "no targets" failure.

Fix this by uploading the populated `requests-cache` filesystem cache (`/tmp/bakery_cache/`) as a run-scoped artifact after the matrix step, then restoring it before each `bakery build` and `bakery ci publish` invocation. The cache keys are deterministic (blake2b of the normalized request URL), so the pre-seeded responses are found by lookup without any code changes to bakery itself. The restore step uses `continue-on-error` so runs that build no dev versions (empty cache) degrade gracefully.

Closes #304

### DIFF
--- a/.github/workflows/bakery-build-native.yml
+++ b/.github/workflows/bakery-build-native.yml
@@ -188,6 +188,13 @@ jobs:
           echo "platform_matrix=$(echo "$result" | jq --compact-output .)" >> "$GITHUB_OUTPUT"
           echo "image_matrix=$(echo "$result" | jq --compact-output '[.[].image] | unique')" >> "$GITHUB_OUTPUT"
 
+      - name: Upload CDN cache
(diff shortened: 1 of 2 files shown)


bschwedler:
Superseded by #552, which seeds the same requests-cache via a run-scoped actions/cache in setup-bakery (with a per-invocation key) and raises the TTL for the life of the run.

PR #627, opened 2026-06-18
Set CI=true for Quarto TinyTeX install command

bschwedler:
Quarto renders an animated TinyTeX download progress bar unless its `runningInCI()` check passes. That check only reads environment variables (`CI`, `GITHUB_ACTIONS`, etc.), none of which propagate into a `docker build` RUN step — so inside container builds Quarto thinks it is interactive and redraws the progress bar on every download tick.

In a non-TTY build log each redraw becomes its own line. One observed Workbench dev build emitted ~26,000 of them (~5 MB, 68% of the log), which slows the build and bloats CI logs.

Setting `CI=true` inline on the install command makes `runningInCI()` pass, so Quarto collapses the progress bar to a single line. Unlike re-adding `--quiet` (removed in 51c7fc66 so install errors stay visible), this keeps all error output, and the variable is scoped to the command so it does not leak into the built image.

### DIFF
--- a/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
+++ b/posit-bakery/posit_bakery/config/templating/macros/quarto.j2
@@ -48,12 +48,19 @@ backward compatibility but is ignored.
 
 {# Command to install TinyTeX using the Quarto CLI
 
+`CI=true` is set inline so Quarto's `runningInCI()` check passes inside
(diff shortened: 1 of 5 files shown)


bschwedler:
Merging this to fix long-running workbench dev version builds

Will have follow-up PRs in the individual repos.

PR #641, opened 2026-07-01
feat: parallelize bakery ci publish across image targets

ianpittwood:
## Summary

Stacked on #619 (unmerged) — targets that branch rather than `main` so this diff only shows the new work.

Parallelizes `bakery ci publish`'s per-target pipeline across image targets, while keeping each target's own operation sequence (wait-for-sources → index-create → soci-convert → index-copy → verify) strictly ordered:

- **Stage 1 (parallel):** wait-for-sources + index-create + soci-convert run concurrently, one job per target, via a new `ParallelShellExecutor.run_jobs()`/`ShellJob`/`CommandRunner` (`parallel/executor.py`) built on the same tracked-spawn/Ctrl-C-safety primitive the existing dgoss `run()` path uses. One target's failure is isolated — it no longer aborts the whole run.
- **Stage 2 (sequential, ordered):** index-copy (the actual registry push) stays serialized in `push_sort_key` order, so registries that display tags by push time (e.g. Docker Hub) still show the latest version as most recent.
- **Stage 3 (sequential):** verify, unchanged — read-only and order-independent.
- Adds `--jobs`/`-j` to `bakery ci publish`, matching the existing dgoss flag.

Out of scope (deferred to follow-up work, per the design spec): retry-with-backoff plumbing in the `parallel` module (transient registry errors are already handled independently by `posit_bakery/retry.py`), and unifying `oras merge`/`soci convert` into this same pipeline.

Design spec and implementation plan (local-only, gitignored under `docs/superpowers/`) drove this; happy to share their content in review if useful.

## Test plan

- [x] `uv run pytest test/parallel/ test/plugins/builtin/imagetools/ test/cli/test_ci_publish.py -q` — 174 passed
- [x] `uv run ruff check .` / `uv run ruff format --check .` — clean
- [x] Task-by-task review (spec compliance + code quality) during implementation; whole-branch review across the full diff — no Critical/Important findings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #599, opened 2026-06-11
feat: parallelize ci publish with retry via a unified imagetools plugin

ianpittwood:
Supersedes #598 

## What & why

`bakery ci publish` ran its 4-phase pipeline (oras index-create → soci convert → oras index-copy → verify) serially per target, with no retry — so the transient GHCR eventual-consistency failures in #591 ("not found" / "manifest unknown" on freshly-pushed digests) failed the whole job and only self-healed on a manual re-run.

Building on the `feat/parallel-calls` parallel module (#588), this:

- **Parallelizes publish.** Each target's create → soci → copy → verify sequence runs as one job on the parallel executor; independent targets publish concurrently (bounded workers, live progress table, Ctrl-C-safe process-group termination).
- **Adds retry-with-backoff.** Every registry command retries on transient errors (5 attempts, exponential 2s→32s + jitter); permanent auth/reference errors fail fast. The capability lives in the `parallel` module so future flaky commands can reuse it.
- **Unifies oras + soci into one `imagetools` plugin.** A single plugin owns the whole pipeline and registers `oras merge`, `soci convert`, and `imagetools publish` (command names preserved). `ci publish`/`ci merge` route through it. `tool: soci` options still parse (the tool-options registry now keys by the options class's `tool` discriminator).

## Parallel module additions

`RetryPolicy`, `CommandResult`, `CommandRunner`, `ShellJob`, `JobResult`, `run_jobs()` — added alongside the existing `ShellTask`/`run()` (dgoss) path, sharing one `_spawn_and_communicate` primitive so timeout/termination/interrupt behavior is identical.

## Verification

- `just test` — **1769 passed**; ruff check + format clean.
- Discovery: plugins = {dgoss, hadolint, imagetools, wizcli}; tool options include `soci`.
- CLI smoke: `oras merge`, `soci convert`, `ci publish`, `ci merge`, `imagetools publish` all register.
- A live `bakery ci publish --dry-run` is blocked by local command policy; the BDD `ci merge` scenario + no-mock dry-run integration tests cover the end-to-end path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #642, opened 2026-07-02
feat: retry-with-backoff for the parallel publish path

ianpittwood:
Stacked on #641. Part of splitting #599 (parallelize + retry + unify oras/soci into `imagetools`) into smaller, independently-reviewable pieces.

## What & why

Retry-with-backoff for transient registry errors (`RetryPolicy`/`retry_on_transient` in `posit_bakery/retry.py`) already covered the riskiest phases of the parallel publish path introduced in #641 — index-create and the SOCI pull step, both carried forward from a pre-existing GHCR-flakiness fix. Two gaps remained:

- **Stage 3 (verify) had no retry.** `OrasIndexVerifyWorkflow` did a bare `oras manifest fetch` — a transient read-after-write blip failed the whole publish with zero backoff, even though every phase before it was protected.
- **Retry backoff wasn't interruptible.** `retry_on_transient` slept via plain `time.sleep`. A Stage-1 job (running on a `ParallelShellExecutor` worker thread) blocked mid-backoff wouldn't notice Ctrl-C — the executor's interrupt handling returns immediately on the main thread, but `ThreadPoolExecutor` registers an atexit hook that joins every thread it ever spawned, so the process wouldn't actually exit until that sleep (and any remaining retry attempts) finished, up to ~62s per stuck job with default settings.

## Changes

- `retry_on_transient()` gains an injectable `sleep` callable (defaults to `time.sleep`, looked up fresh per call so existing `patch("posit_bakery.retry.time.sleep")` tests keep working unmodified).
- `CommandRunner.sleep()` (new, in the `parallel` module): sleeps in short slices, checking the bound executor's shutdown flag between slices, raising a new `ExecutorInterrupted` as soon as shutdown is observed — so a mid-backoff job unwinds within about one slice interval of Ctrl-C instead of blocking process exit.
- `OrasIndexCreateWorkflow` and `SociConvertWorkflow`'s pull-step retries (the only retries reachable from a Stage-1 `ShellJob` worker thread) now pass `sleep=runner.sleep` when a runner is available.
- `OrasIndexVerifyWorkflow` gains a `retry_policy` field and now retries transient errors on the final tag-resolution check, mirroring the existing Create/Copy pattern. No `runner`/interrupt plumbing there — it runs on the main thread, not inside the executor, so there's no process-exit-blocking risk to fix.

Explicitly out of scope (see design doc): unifying `OrasWaitForSourcesWorkflow`'s poll loop onto `RetryPolicy` (different, already-correct mechanism), adding jitter, changing `RetryPolicy` defaults.

## Verification

- `just test` — 1947 passed; `ruff check` + `ruff format --check` clean.
- Each commit individually reviewed (spec compliance + code quality) before merge into this branch; a final whole-branch review confirmed the cross-task composition (retry → `CommandRunner.sleep` → `ExecutorInterrupted` propagation through `_run_job`) is correct end to end and all stated non-goals were honored.
- Known follow-up (not blocking): `OrasWaitForSourcesWorkflow`'s own poll sleep is the one Stage-1 wait left non-interruptible (~5s Ctrl-C latency bound vs. the ~62s this PR fixes for retry backoff) — same one-line fix pattern would close it if desired.

### DIFF
--- a/posit-bakery/posit_bakery/parallel/__init__.py
+++ b/posit-bakery/posit_bakery/parallel/__init__.py
@@ -1,5 +1,6 @@
 from posit_bakery.parallel.executor import (
     CommandRunner,
+    ExecutorInterrupted,
     JobResult,
     ParallelShellExecutor,
     ShellJob,
@@ -10,6 +11,7 @@ from posit_bakery.parallel.executor import (
 
 __all__ = [
     "CommandRunner",
+    "ExecutorInterrupted",
     "JobResult",
     "ParallelShellExecutor",
     "ShellJob",
--- a/posit-bakery/posit_bakery/parallel/executor.py
+++ b/posit-bakery/posit_bakery/parallel/executor.py
@@ -22,6 +22,20 @@ from posit_bakery.settings import SETTINGS
 # its container via its EXIT trap — before we escalate to SIGKILL.
 TERMINATE_GRACE_SECONDS = 10.0
 
+# Slice width for CommandRunner.sleep()'s interruptible backoff wait. Short
+# enough that a shutdown request is noticed promptly without busy-waiting.
+RETRY_SLEEP_SLICE_SECONDS = 0.5
+
+
+class ExecutorInterrupted(Exception):
+    """Raised by :meth:`CommandRunner.sleep` when the bound executor's shutdown
+    flag flips mid-wait.
+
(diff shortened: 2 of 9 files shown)


ianpittwood:
Tracked by #648.

bschwedler (approved):

PR #655, opened 2026-07-15
Guard release PR creation against duplicate dispatches

### LINKED ISSUE #653: Release workflow can collide with an in-flight release PR
images-workbench's Release workflow failed when the shared product-release.yml force-pushed to release/2026.07.0. GitHub rejected the push (GH006) because a PR for that branch had already been added to a merge queue, and queued branches can't be updated.

The run had been dispatched a second time for the exact same version (2026.07.0+139.pro9) that an earlier, successful run had already used to open PR #163, which was already progressing toward merge.

product-release.yml always force-pushes to the release branch unconditionally, with no regard for the upstream branch's state. Any re-dispatch while a prior release PR is in flight fails this way — after bakery.yaml has already been mutated and committed locally, not before.

This workflow is shared by all three product repos (images-connect, images-package-manager, images-workbench), so the same collision can happen to any of them.

- https://github.com/posit-dev/images-workbench/actions/runs/29354432682

bschwedler:
product-release.yml always force-pushed to the release branch, even when a prior dispatch's PR was already in flight or already merged — colliding with merge-queued branches (GH006) or duplicating already-shipped releases.

Add a no-op check before the commit/push: skip when nothing is staged, or when the staged content already matches what's on the release branch. Genuine changes still commit and force-push as before.

### DIFF
--- a/.github/workflows/product-release.yml
+++ b/.github/workflows/product-release.yml
@@ -153,8 +153,29 @@ jobs:
           git add -A
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git commit -m "Release ${DISPLAY_VERSION}"
-          git push -u origin "$BRANCH" --force-with-lease
+
+          if git diff --cached --quiet; then
+            # Nothing changed vs. HEAD: this display version's content is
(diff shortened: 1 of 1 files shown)


ianpittwood (approved):
Makes sense to me!

PR #662, opened 2026-07-15
feat: parallelize --strategy build image builds

### LINKED ISSUE #647: feat: parallelize serial (--strategy build) image builds
## Summary

`BakeryConfig.build_targets()` (`posit-bakery/posit_bakery/config/config.py`) has an `ImageBuildStrategy.BUILD` branch that loops over `self.targets` and calls `target.build(...)` serially, wrapped in `_retry_build`. `--strategy build` exists specifically for generic container builders (Podman, nerdctl) that don't support Buildx Bake's native parallelism, so it currently gets none of the concurrency that `--strategy bake` gets for free.

Unlike hadolint/wizcli (see related issues), this path does **not** shell out via a plain `cmd: list[str]` — `target.build()` (`posit_bakery/image/image_target.py`) calls the `python_on_whales` SDK directly. The existing `ShellTask`/`ParallelShellExecutor` primitives from #588 assume a raw subprocess command, so this isn't a drop-in port.

## Scope (needs a design decision before implementation)

- Investigate whether to:
  1. Extend `posit_bakery/parallel/` with a callable-based job abstraction (run an arbitrary function per target on the executor, not just a shell command) — generalizing the direction `ShellJob`/`CommandRunner` took in #641/#642 for oras/soci.
  2. Or shell out to `docker build`/`podman build` directly as a `cmd: list[str]`, dropping down from the `python_on_whales` SDK for this strategy specifically.
- Bound concurrency and preserve existing `--fail-fast`, `retry`, and `BakeryBuildErrorGroup` aggregation behavior.
- Preserve `_merge_sequential_build_metadata_files()` metadata-file writing behavior when `--metadata-file` is set.

## Out of scope

- Changing `--strategy bake` behavior (already parallel via Buildx Bake).

Part of the investigation started in #584.

ianpittwood:
## Summary

Closes #647. `bakery build --strategy build` (used for generic container builders like Podman that can't use Buildx Bake) built targets serially through `python_on_whales`. This makes it build concurrently, matching `--strategy bake`'s parallelism, while preserving retry/`--fail-fast`/metadata-file behavior.

- Made `ImageTarget.build()` cwd-independent (dropped a process-wide `contextlib.chdir` that was not thread-safe) and gave it an optional `log_callback` to stream output line-by-line instead of returning an `Image`.
- Added `PrefixedLogSink`, a thread-safe console sink, so concurrent targets' streamed output stays readable with a `[uid]` prefix per line (went through two fix rounds to correctly avoid Rich markup parsing mangling bracketed docker/buildx output like `[internal]`).
- Added `fail_fast` to `ParallelShellExecutor.run()`/`run_jobs()`: on the first failure, not-yet-started jobs are cancelled while already-running ones finish and are still collected.
- Rewrote `BakeryConfig.build_targets()`'s `BUILD` branch to fan out over `ShellJob`s via the existing `run_jobs()` machinery (reused from #641/#642's parallel publish work) instead of building a new executor, wiring the above together. Added a `jobs` parameter.
- Added a `--jobs`/`-j` CLI flag to `bakery build`, consistent with `dgoss run`/`imagetools publish`/`ci matrix`.

Design doc: `docs/superpowers/specs/2026-07-15-parallel-build-strategy-design.md` (local, gitignored).

Built and reviewed task-by-task (implementer + task reviewer per task, plus a final whole-branch review) via subagent-driven-development; all findings from that process were fixed and re-verified before this PR.

## Test plan

- [x] `test/image/test_image_target.py` — cwd-independence + log-streaming coverage
- [x] `test/parallel/test_executor.py` — `PrefixedLogSink` (incl. bracketed-content regression test) + `fail_fast` cancellation behavior
- [x] `test/config/test_config.py` — new `TestBuildTargetsBuildStrategy` covers concurrency bound, single/multi-failure error aggregation, fail-fast, metadata file, and an end-to-end test exercising the real streaming wiring (no mocked `ImageTarget.build`)
- [x] `test/cli/test_build.py` — `--jobs` flag plumbing
- [x] Full suite green: 1968 tests passing (`test/cli/` run separately due to a pre-existing, unrelated cwd-pollution issue when run in the same process as the rest of the suite)


ianpittwood:
Tested out in a PR on https://github.com/posit-dev/images-workbench/pull/167

ianpittwood:
Pulled the latest changes (your version fix is now in place) and compared the two runs.

**Overall run wall-clock:**
| Run                    | Duration        |
| ---------------------- | --------------- |
| Serial (29502072150)   | 22m 18s         |
| Parallel (29504536583) | 13m 58s         |
| **Reduction**          | **~37% faster** |

**Per-image build/test job comparison** (matched by job name across both runs, 86 jobs total):

| Metric                   | Serial    | Parallel  | Change |
| ------------------------ | --------- | --------- | ------ |
| Sum of all job durations | 424.6 min | 362.8 min | −14.6% |
| Average job duration     | 296s      | 253s      | −14.6% |

The overall wall-clock win (37%) is much bigger than the average per-job win (14.6%) because the bottleneck jobs — the ones that were gating the whole workflow — shrank the most:

| Job                                                      | Serial | Parallel | Reduction |
| -------------------------------------------------------- | ------ | -------- | --------- |
| Production `workbench:2026.07.0+139.pro9 (amd64)`        | 1225s  | 743s     | **39.3%** |
| Production `workbench:2026.05.1+225.pro10 (amd64)`       | 945s   | 627s     | **33.7%** |
| Session `workbench-session:R4.3.3-python3.12.13 (amd64)` | 524s   | 253s     | **51.7%** |
| Session `workbench-session:R4.5.3-python3.14.6 (amd64)`  | 505s   | 253s     | **49.9%** |
| Production `workbench:2026.06.0+242.pro13 (amd64)`       | 822s   | 610s     | **25.8%** |
| Development `workbench:2026.08.1-daily+53.pro14 (amd64)` | 681s   | 471s     | **30.8%** |

Most `amd64` jobs (which were the longest-running/critical-path jobs) saw 25–52% reductions, while `arm64` jobs — already shorter — saw smaller gains (5–10%). Net effect: the Production job group's wall-clock span dropped from 1296s → 743s (43% faster), which was the biggest driver of the overall run-time improvement.

**Confirmed: parallel build calls measurably reduce per-image build/test time, especially for the slower `amd64` builds that were previously the critical path.**

posit-bakery/posit_bakery/config/config.py
```diff
@@ -1169,27 +1175,37 @@ def build_targets(
...
-            for target in self.targets:
-                try:
-                    _retry_build(
-                        lambda t=target: t.build(
+            sink = PrefixedLogSink()
+            shell_jobs = [
+                ShellJob(
+                    key=target.uid,
+                    label=str(target),
+                    run=lambda runner, t=target: _retry_build(
```
bschwedler, line 1193: **Blocking:** this job ignores `runner` and calls `python_on_whales.docker.build()` directly, so nothing in the build path responds to shutdown:

- `_retry_build`'s `time.sleep` (config.py:72) blocks uninterruptibly — under `--fail-fast`/SIGTERM a sibling burns its full retry delay before stopping.
- an in-flight build isn't in `_active`, so `_terminate_active()` can't kill it; Ctrl-C only works via tty process-group luck.

Fix (retry half): make `_shutdown` a `threading.Event`; `_retry_build` waits on `stop.wait(_RETRY_DELAY_SECONDS)` — aborts instantly, and lets `CommandRunner.sleep` drop its polling.

Residual (won't fix): python_on_whales owns the build `Popen` and doesn't expose it (utils.py:298), so a mid-build target can't be killed without reimplementing the spawn. Suggest documenting.

posit-bakery/posit_bakery/config/config.py
```diff
@@ -1169,27 +1175,37 @@ def build_targets(
...
+                        lambda: t.build(
                             load=load,
                             push=push,
                             pull=pull,
                             cache=cache,
                             platforms=platforms,
                             metadata_file=True if metadata_file else None,
+                            log_callback=lambda line, u=t.uid: sink.write(u, line),
```
bschwedler: **Required:** `log_callback` is set unconditionally, so `stream_logs` is always on, forcing `progress="plain"` (image_target.py:686) and streaming every build line regardless of `SETTINGS.log_level`. `-q` no longer quiets `--strategy build` — silent regression, no test.

Fix: gate streaming on log level, or drop lines in the sink when quiet.

bschwedler (changes requested):
Review feedback assisted by the [critical-code-reviewer skill](https://github.com/posit-dev/skills/blob/main/posit-dev/critical-code-reviewer/SKILL.md).

posit-bakery/posit_bakery/config/config.py
ianpittwood: Fixed.

posit-bakery/posit_bakery/config/config.py
ianpittwood: `_retry_build` now accepts a sleep callable, wired to `runner.sleep` (which already polls for shutdown) instead of a bare `time.sleep`.

bschwedler (approved):

PR #663, opened 2026-07-16
Make push-to-main builds change-aware

### LINKED ISSUE #652: Release merges rebuild every version, not just what changed
Merging a release PR into a product repo (images-connect, images-package-manager, images-workbench) triggers production.yml on push to main, which builds every non-dev version of every image in that repo — not just the version(s) the release touched.

Merging PR #163 ("Release 2026.07.0" in images-workbench) only changed files under workbench/2026.07/ and workbench-session-init/2026.07/, but the resulting run built 6 versions across 2 platforms (21 jobs), including versions untouched since earlier releases.

production.yml's build job runs the same full-matrix build on every push to main as it does on the weekly scheduled rebuild. It doesn't distinguish the two triggers or look at what actually changed.

This wastes CI compute and time on every release merge, and the cost grows as more versions accumulate in bakery.yaml.

- https://github.com/posit-dev/images-workbench/actions/runs/29355616351/job/87168719890
- https://github.com/posit-dev/images-workbench/pull/163

bschwedler:
Release merges rebuild every version in the matrix regardless of what
changed, because the build workflows never pass `--base-ref` to `bakery
ci matrix` on push events — only PR builds get that treatment today.
Merging a release PR that touched 2 version directories still triggers
a 21-job full-matrix build across 6 versions, on every release.

This reuses the existing PR-side change-classification machinery for
push-to-main events too, diffing against `github.event.before` the
same way PR builds diff against their base commit. Two related gaps
surfaced during review and are fixed here as well:

- A git-diff failure (unreachable ref, e.g. the all-zero SHA on a
  branch's first push) used to crash the build; it now falls back to
  a full build with a warning.
- An empty change-aware matrix (e.g. a docs-only push) used to fail
  the build outright, since GitHub Actions fails rather than skips a
  job fed an empty matrix; both workflows now guard against this the
  same way the PR workflow already does.

`schedule` and `workflow_dispatch` triggers are unaffected and
continue to always build the full matrix.

`bakery.yaml` changes were still an unconditional full-matrix
fail-safe, which defeated the above for the exact case it targets:
`bakery create version` always writes the new version into
`bakery.yaml`, so every release merge tripped that fail-safe
regardless of the push-to-main change-awareness above. `bakery ci
matrix` now structurally diffs `bakery.yaml`'s old and new content
(parsed YAML, not text) and narrows to the specific image(s)/
version(s) actually affected, falling back to a full build only when
the diff can't be confidently understood (unparseable YAML, malformed
shape, or a change outside `images[]`).

- Closes https://github.com/posit-dev/images-shared/issues/652

### DIFF
--- a/posit-bakery/posit_bakery/cli/ci.py
+++ b/posit-bakery/posit_bakery/cli/ci.py
@@ -44,7 +44,9 @@ def _resolve_changed_files(base_ref: str | None, changed_files_from: str | None,
 
     - ``--base-ref`` runs ``git diff`` (paths relative to the repo root), then
       rebases them onto ``rebase_root`` (the bakery context) and drops paths outside
-      it.
+      it. If the diff against ``base_ref`` fails for any reason (unreachable ref,
+      all-zero SHA, unrelated histories), logs a warning and returns ``None``
+      (disabling filtering) rather than raising.
     - ``--changed-files-from`` paths are used verbatim and must already be relative
       to the bakery context root. They are not rebased, so do not pipe raw
       ``git diff --name-only`` output here unless the context is the repo root. Use
@@ -72,8 +74,22 @@ def _resolve_changed_files(base_ref: str | None, changed_files_from: str | None,
         text=True,
     ).stdout.strip()
     repo_root = Path(toplevel)
+
+    try:
+        changed = git_changed_files(repo_root, base_ref)
+    except subprocess.CalledProcessError as e:
+        # base_ref may not diff cleanly (e.g. github.event.before is the all-zero SHA
+        # on a branch's first push, or an unreachable ref after a force-push). Fail
+        # safe to a full build rather than crashing the CI run.
+        log.warning(
+            "Could not diff against base-ref %r (%s); falling back to a full build.",
+            base_ref,
+            e,
+        )
+        return None
+
     rebased: list[str] = []
-    for rel in git_changed_files(repo_root, base_ref):
+    for rel in changed:
         abs_path = repo_root / rel
         try:
             rebased.append((abs_path.relative_to(rebase_root)).as_posix())
--- a/posit-bakery/posit_bakery/cli/ci.py
+++ b/posit-bakery/posit_bakery/cli/ci.py
@@ -71,13 +71,7 @@ def _resolve_changed_files(base_ref: str | None, changed_files_from: str | None,
     # isn't a git checkout at all, a more fundamental misconfiguration than a
     # base_ref that doesn't diff cleanly, and should surface loudly rather than
     # silently fall back to a full build.
-    toplevel = subprocess.run(
-        ["git", "-C", str(rebase_root), "rev-parse", "--show-toplevel"],
-        check=True,
-        capture_output=True,
-        text=True,
-    ).stdout.strip()
-    repo_root = Path(toplevel)
+    repo_root = _git_repo_root(rebase_root)
 
     try:
         changed = git_changed_files(repo_root, base_ref)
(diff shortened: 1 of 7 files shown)


bschwedler:
Push-triggered builds previously always built the full matrix. This reuses the narrowing logic PR builds already had, but keyed on a different reference point: PR builds diff against the PR's base commit (`github.event.pull_request.base.sha`), while a push has no PR to anchor to — so it diffs against `github.event.before`, the commit that was `main`'s tip immediately before this push landed. That's the direct analog of a PR's base commit for a push event.

Two new safety nets cover cases PR usage didn't need to handle:

- **Fail-safe**: if the diff against `BASE_REF` can't be computed at all (e.g. the all-zero SHA GitHub sends as `github.event.before` on a branch's first-ever push), log a warning and fall back to a full build instead of crashing.
- **Empty-matrix guard**: if the diff succeeds but resolves to nothing buildable (a docs-only merge), the resulting matrix is legitimately empty — but GitHub Actions fails, not skips, a job fed an empty matrix, so the affected jobs are now explicitly guarded to skip instead.

`schedule`/`workflow_dispatch` are untouched — `github.event.before` doesn't exist on those event payloads.

```mermaid
flowchart TD
    Trigger{Trigger event} -->|push| SetRef["BASE_REF = github.event.before"]
    Trigger -->|"schedule / workflow_dispatch"| NoRef["BASE_REF unset, unchanged"]

    NoRef --> FullBuild["Full matrix:<br/>every image x version"]

    SetRef --> Diff{"Does the git diff<br/>against BASE_REF succeed?"}

    Diff -->|"No - bad ref, all-zero SHA<br/>on a first push, unrelated histories"| FailSafe["ci.py logs a warning,<br/>disables filtering"]
    FailSafe --> FullBuild

    Diff -->|Yes| Classify{"What did the diff touch?"}

    Classify -->|".github/workflows/**"| FullBuild
    Classify -->|"bakery.yaml"| YamlDiff{"Can old bakery.yaml be read<br/>and confidently diffed<br/>against the new?"}
    Classify -->|"specific images/versions"| PartialBuild["Partial matrix:<br/>only what changed"]
    Classify -->|"nothing buildable,<br/>e.g. docs-only push"| EmptyMatrix["Empty matrix"]

    YamlDiff -->|"No - unreadable, unparseable,<br/>malformed, or outside images[]"| FullBuild
    YamlDiff -->|"Yes - narrows the same way<br/>file-path changes do"| PartialBuild

    FullBuild --> RunNormally["build-test/build then merge then readme<br/>run normally"]
    PartialBuild --> RunNormally

    EmptyMatrix --> Guard{"Is the matrix non-empty?"}
    Guard -->|No| SkipAll["build-test/build: skipped<br/>merge: skipped, cascades<br/>readme: skipped, cascades"]
```

bschwedler:
@ianpittwood I've switched this over to using the Artifact.

posit-bakery/posit_bakery/cli/ci.py
```diff
@@ -223,7 +288,13 @@ def matrix(
         selection: MatrixSelection | None = None
         changed = _resolve_changed_files(base_ref, changed_files_from, c.base_path)
         if changed is not None:
-            selection = classify_changes(c, changed)
+            old_bakery_yaml = _resolve_base_bakery_yaml(base_ref, changed, c.config_file)
```
ianpittwood: `--changed-files-from` does not fully override `--base-ref` as it would still be passed here when both are set. Seems we either need to compute the `base_ref` or modify the warning.

ianpittwood (approved):
The rest looks good. Just one spot that seems like it could have undesirable behavior.

posit-bakery/posit_bakery/cli/ci.py
bschwedler: Fixed in e3cd5952 — `_resolve_base_bakery_yaml` now gets `None` for `base_ref` whenever `--changed-files-from` is set, so the bakery.yaml semantic diff no longer runs git against `base_ref` when it's supposed to be overridden. Added a regression test that fails on the old behavior (silently narrows using the leaked `base_ref`) and passes now (falls back to the full matrix).

PR #708, opened 2026-08-06
Add restore-cache opt-out to setup-bakery

### LINKED ISSUE #686: Eliminate error messages from missing cache artifacts on first run
On the first run of a workflow (or when the artifact cache is missing), the `Restore dependency version cache` step in `setup-bakery` attempts to download a non-existent artifact, producing error messages like:
```
Unable to download artifact(s): Artifact not found for name: bakery-deps-*
Please ensure that your artifact is not expired and the artifact ...
```

Although the workflow continues due to `continue-on-error: true`, these error messages appear as failed steps in job logs, creating noise and confusion.

## Expected behavior

The cache restore should fail gracefully on first run without producing visible error messages, while still showing actual retrieval failures (permissions, expiration, network issues).

## Possible approach

Query the GitHub API in a bash step to check if the expected cache artifact exists before attempting to download it, conditionally running the `download-artifact` action only when the artifact exists.

- [ ] https://github.com/posit-dev/images-shared/pull/708
- [ ] https://github.com/posit-dev/images-connect/pull/146
- [ ] https://github.com/posit-dev/images-package-manager/pull/132
- [ ] https://github.com/posit-dev/images-workbench/pull/178
- [ ] https://github.com/posit-dev/images-examples/pull/43

bschwedler:
Several setup-bakery callers never run alongside an is-writer job in
their invocation, so the "Restore dependency version cache" step's
download-artifact call always misses -- not a cold-cache edge case,
but every single run. continue-on-error hides the resulting job
failure, but the noisy "Artifact not found" error still clutters
logs.

Add a restore-cache input (default "true") that lets those callers
skip the restore step entirely instead of attempting a download that
can never succeed, and wire it to "false" on clean.yml, hadolint.yml,
and product-release.yml here.

This is narrower than the API-existence-check originally proposed in
the issue below: it silences the guaranteed-miss cases but doesn't
stop continue-on-error from masking a genuine restore failure for
jobs that do participate in the shared cache. Leaving that open as a
possible follow-up rather than folding it into this change.

images-connect, images-package-manager, images-workbench, and
images-examples each have a matching one-line change queued, blocked
on this merging to main since they pin setup-bakery@main and
composite actions reject unrecognized inputs.

- Closes https://github.com/posit-dev/images-shared/issues/686

### DIFF
--- a/.github/workflows/clean.yml
+++ b/.github/workflows/clean.yml
@@ -99,6 +99,7 @@ jobs:
         uses: "posit-dev/images-shared/setup-bakery@main"
         with:
           version: ${{ inputs.version }}
+          restore-cache: "false"
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee  # v4.2.0
@@ -144,6 +145,7 @@ jobs:
         uses: "posit-dev/images-shared/setup-bakery@main"
         with:
           version: ${{ inputs.version }}
+          restore-cache: "false"
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee  # v4.2.0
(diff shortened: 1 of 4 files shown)


ianpittwood (approved):
LGTM

Issue #709, opened 2026-08-06
Add `bakery build --summary` for build/tag/size counts


bschwedler:
`bakery build` fans out into dozens of image builds and hundreds of pushed tags per product repo, but nothing reports the shape or cost of that work. Today the only way to find out is hand-counting `--plan` JSON output. Measured live across the three product repos:

| Repo | Bake targets | Registry tags | Platform builds |
|---|---|---|---|
| images-connect | 53 | 272 | 64 |
| images-workbench | 30 | 286 | 39 |
| images-package-manager | 24 | 180 | 32 |

Add a `--summary` flag to report counts (and, eventually, image/cache sizes) for whatever the current filters selected, as a human-readable table or JSON for CI.

This ships in phases:

- [x] **Phase 1 (counts, no I/O)** — build target, platform build, and registry tag counts, derived from already-resolved config so every existing filter flag works for free.
- [x] **Phase 2 (image sizes)** — registry (compressed) and local (uncompressed) size columns, populated after a real build via `python_on_whales`.
- [x] **Phase 3 (cache size)** — layer cache size per target, needs a new read-only GHCR registry v2 client (`ghcr.io`), since the registry Packages API (`api.github.com`) bakery currently talks to exposes no size field at all.
- [x] https://github.com/posit-dev/images-shared/pull/710
- [x] https://github.com/posit-dev/images-shared/pull/711
- [x] https://github.com/posit-dev/images-shared/pull/712

PR #710, opened 2026-08-06
Add bakery build --summary for target/tag counts

bschwedler:
This ships Phase 1 of the `--summary` design: build target, platform build, and registry tag counts for whatever the current filters selected, with no build or registry I/O. Phases 2 (image sizes) and 3 (cache size) follow separately, since they need a real build and a registry client bakery doesn't have yet.

- `--summary-format {table,json}` defaults to table; table goes to stderr, json to stdout, matching the existing stream convention.
- `--summary-format json` is a hard error with `--plan`, mirroring the existing `--plan` + `--strategy build` guard, since `--plan` already owns stdout JSON.
- Registry tag counts are aliases (version, os, latest, etc.) across registries, not distinct stored images, so the table caption calls that out explicitly.

Verified live against images-connect, images-workbench, and images-package-manager via `--context`: counts match hand-counted figures exactly, and `--latest` correctly narrows filtered counts.

https://github.com/posit-dev/images-shared/issues/709

### DIFF
--- a/posit-bakery/posit_bakery/cli/build.py
+++ b/posit-bakery/posit_bakery/cli/build.py
@@ -14,9 +14,9 @@ from posit_bakery.config.image.posit_product.errors import (
     ArtifactNotAvailableError,
     VersionSubstitutionError,
 )
-from posit_bakery.const import DevVersionInclusionEnum, MatrixVersionInclusionEnum
+from posit_bakery.const import DevVersionInclusionEnum, MatrixVersionInclusionEnum, SummaryOutputFormat
 from posit_bakery.error import BakeryBuildErrorGroup, BakeryToolRuntimeError
-from posit_bakery.image import ImageBuildStrategy
+from posit_bakery.image import BuildSummary, ImageBuildStrategy
 from posit_bakery.log import stderr_console, stdout_console
 from posit_bakery.util import auto_path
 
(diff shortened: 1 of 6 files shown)


posit-bakery/posit_bakery/cli/build.py
```diff
@@ -287,7 +303,17 @@ def build(
...
+                "❌ --summary-format json is not supported with --plan, which already writes "
+                "the bake plan as JSON to stdout. Use --summary-format table, or drop --plan.",
+                style="error",
+            )
+            raise typer.Exit(code=1)
         stdout_console.print_json(config.bake_plan_targets(push=push))
+        if summary:
+            build_summary = BuildSummary.from_image_targets(config.targets, platforms=image_platform)
+            stderr_console.print(build_summary.table(sizes=False))
         raise typer.Exit(code=0)
```
ianpittwood: I feel like this shouldn't be contingent on `--plan`? I realize it's confusing having both, but having `--summary` nested under `--plan` with no note is also confusing as a first time user.

ianpittwood (commented):
I'd like to clarify the expected option behavior before moving any further.

posit-bakery/posit_bakery/cli/build.py
bschwedler: Good point. Let me consolidate the logic in a better way.

bschwedler: I've combined down the summary logic into a single location.

posit-bakery/posit_bakery/image/summary.py
```diff
@@ -0,0 +1,81 @@
...
+            else (len(target.image_os.platforms) if target.image_os else len(DEFAULT_PLATFORMS))
+            for target in targets
+        )
+        registry_tags = sum(len(target.tags) for target in targets)
+
+        return cls(
+            rows=[
+                BuildSummaryRow(key="build_targets", label="Build Targets", value=len(targets)),
+                BuildSummaryRow(key="platform_builds", label="Platform Builds", value=platform_builds),
+                BuildSummaryRow(key="registry_tags", label="Registry Tags", value=registry_tags),
```
ianpittwood, line 54: We may want to refine the terms used here?
- Unique Images?
- Manifests?

ianpittwood (approved):
LGTM, manually tested.

PR #711, opened 2026-08-06
Add image size and layer counts to build --summary (Phase 2)

bschwedler:
This PR targets `feat/build-summary-flag` (Phase 1, not yet merged), not `main` — it's stacked on top of it and should merge after.

Phase 2 of the `--summary` design: registry (compressed) and local (uncompressed) image size, plus a layer count, alongside Phase 1's counts. Both size lookups are best-effort — a measurement failure never fails the build, it just leaves that value unmeasured.

- Once sizes are available, the table switches from Phase 1's 3-row aggregate view to a per-target breakdown (grouped by image/version/os/variant, mirroring the Goss test report's table), plus a Total row. `--plan` (no sizes possible) is unchanged.
- `--summary` now also prints on the failure paths, not just success — a partially failed build is exactly when the report is most useful.
- A multi-platform registry inspect costs 1+N round trips (the index lists child descriptors, not layers), so it runs through the existing `ParallelShellExecutor`.
- Extracted `grouped_table()` as shared report-table infrastructure and moved the Goss test report onto it, rather than duplicating the same grouping/Total-row algorithm a second time.

- https://github.com/posit-dev/images-shared/pull/710
- https://github.com/posit-dev/images-shared/issues/709


posit-bakery/posit_bakery/image/summary.py
```diff
@@ -1,15 +1,132 @@
...
+    return Text(DASH, style="bright_black italic")
+
+
+def _docker_cmd() -> list[str]:
+    return [str(part) for part in python_on_whales.docker.docker_cmd]
+
+
+def _inspect_local(runner: CommandRunner, ref: str) -> ImageInspectResult | None:
+    """Runs `docker image inspect <ref>` against the local daemon; `None` on any failure."""
+    result = runner.run([*_docker_cmd(), "image", "inspect", ref])
```
ianpittwood: Readd why this is invoked this way instead of through python-on-whales. We should eventually upstream a better implementation there.

posit-bakery/posit_bakery/plugins/builtin/dgoss/report.py
ianpittwood, line 1: In the future let's try to split this sort of refactor stuff off separately.

ianpittwood (approved):
Minor notes, LGTM. Ran as expected manually.

ianpittwood:
@bschwedler looks like there is a test case failure to fix here

posit-bakery/posit_bakery/image/summary.py
bschwedler: I've switched this over to python-on-whales for local.

PR #712, opened 2026-08-07
Add cache size to build --summary (Phase 3)

bschwedler:
This PR targets `feat/build-summary-sizes` (Phase 2, not yet merged), not `main` — it's stacked on top of it and should merge after.

Phase 3 of the `--summary` design: per-target build cache size. The GHCR Packages API bakery already talks to exposes no size field on a package version at all, so this adds a focused read-only client for the registry v2 API (`ghcr.io`), which does.

- Cache measurement is gated on `push` alone, not `load` — `cache_to` is only ever written when pushing, so a local-only build always shows a dash rather than a misleading value.
- A true multi-platform target's cache tag carries no arch suffix, so more than one row can share the same cache ref; totals dedupe by ref before summing so a shared cache is never double-counted.
- The plan's "local-first shortcut" (reading BuildKit's own cache store via `docker buildx du` when `--summary` runs in the same job as the build) isn't implemented here: `buildx du` reports cache by content digest with no link back to a target's tag, so it can't populate a per-target column without guessing.

- https://github.com/posit-dev/images-shared/pull/711
- https://github.com/posit-dev/images-shared/issues/709


posit-bakery/posit_bakery/registry_management/ghcr/manifest.py
ianpittwood, line 1: Is this reliant on using a GHCR backend for caching?

bschwedler: I am switching this over to pulling this information from the manifest.

bschwedler: I switched to `docker buildx imagetools inspect --raw` instead of the new GHCR client, a call which is made elsewhere.

This works with any `--cache-registry`

bschwedler:
Cache size now uses `docker buildx imagetools inspect --raw` instead of the new GHCR client. Same call `summary.py` already makes for image sizes, and it returns the same manifest — checked against `ghcr.io/posit-dev/connect/cache:2026.01.1-ubuntu-24.04-min-amd64`.

The client only worked for ghcr.io and only with a `GITHUB_TOKEN`. Anything else showed a dash. This uses the builder's existing login, so any `--cache-registry` works. `GHCRManifestClient` and its tests are removed; a test with an ECR ref covers the rest.

ianpittwood (approved):

PR #714, opened 2026-08-07
Fix checksum and URL bugs in setup-goss/hadolint

### LINKED ISSUE #713: setup-hadolint/setup-goss: checksum and URL bugs for pinned versions
setup-hadolint and setup-goss each have a dormant bug. A review of oven/Containerfile on branch feat/bakery-oven found both.

**setup-hadolint**: GitHub matches download URLs without case. So the fallback from `hadolint-linux-${arch}` to `hadolint-Linux-${arch}` never runs, because the lowercase request always succeeds. For hadolint versions before v2.13.0, the script saves the binary under the wrong filename. Then the checksum check fails.

**setup-goss**: the "Install dgoss" step builds the URL as `releases/${version}/download/dgoss`. GitHub's real pattern is `releases/download/${tag}/dgoss`. The old URL works only when `version` is `"latest"`. Any pinned version returns a 404 error.

Both bugs are dormant now. Every caller uses the default value `"latest"` for both actions. If anyone pins an older version, the bugs will occur.

bschwedler:
A review of oven/Containerfile on branch feat/bakery-oven found two dormant bugs. Both actions use the default `"latest"` version today, so neither bug shows up yet. If someone pins an older version, each bug will occur.

- setup-hadolint: GitHub matches download URLs without case, so the fallback from a lowercase name to a capitalized name never runs. I confirmed this against hadolint v2.12.0. The fix reads the true filename from the checksum file instead of guessing it.
- setup-goss: the dgoss download URL had the wrong segment order. It worked only for the literal value `"latest"`, by coincidence. I confirmed this against four real goss tags. The fix passes the resolved tag to the dgoss step through `GITHUB_OUTPUT`.

Closes https://github.com/posit-dev/images-shared/issues/713

### DIFF
--- a/setup-goss/action.yml
+++ b/setup-goss/action.yml
@@ -20,6 +20,7 @@ runs:
       shell: bash
       run: mkdir -p tools
     - name: Install goss
+      id: install-goss
       shell: bash
       run: |
         # Determine architecture if not provided
@@ -52,6 +53,7 @@ runs:
           tag="${{ inputs.version }}"
         fi
         version="${tag#v}"
+        echo "tag=${tag}" >> "$GITHUB_OUTPUT"
 
         release_url="https://github.com/goss-org/goss/releases/download/${tag}"
         asset="goss_${version}_linux_${goss_arch}.tar.gz"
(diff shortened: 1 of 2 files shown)


ianpittwood (approved):
LGTM if test rerun fixes flakiness

Issue #482, opened 2026-04-23
Enable security scanning in CI workflows


bschwedler:
Scanning with the `wizcli` was added in:
- https://github.com/posit-dev/images-shared/issues/217

We need to enable this in the workflow once the policies are worked out.

ianpittwood:
I'm currently thinking build-time scans for PRs. Everything else should be adequately monitored by registry scans.

bschwedler:
For now, we will add `wizcli` to the builds on `main` only.

Issue #217, opened 2025-09-22
Implement image security scans using `wizcli`


bschwedler:
Add security scanning with `wizcli` to bakery CLI

PR #715, opened 2026-08-07
Add wizcli security scan to bakery-build-native.yml

bschwedler:
Adds `bakery wizcli scan` (Wiz container vulnerability scanning) as a new advisory-only step in the shared `bakery-build-native.yml` workflow, so every real build (push/schedule/main-dispatch) across `images-connect`, `images-workbench`, and `images-package-manager` gets scanned once wired up downstream.

Auth, policy ID, and project ID all come from four new optional secrets declared on this workflow (`WIZ_CLIENT_ID`/`WIZ_CLIENT_SECRET`/`WIZ_POLICY_ID`/`WIZ_PROJECT_ID`) — `bakery.yaml` is never involved, and repos without the secrets configured skip the step cleanly rather than showing a permanently red one. The Scan step is `continue-on-error: true` throughout: a policy violation is visible in the log but never blocks a build.

This is the first of four coordinated PRs. The other three (`images-connect`, `images-workbench`, `images-package-manager`) forward the new secrets to this workflow and can't merge until this one does — GitHub Actions fails workflow parsing if a caller passes a secret the reusable workflow doesn't declare. Populating the secret values themselves is a separate manual step per repo, not part of any of these PRs.

Related:
- https://github.com/posit-dev/images-shared/issues/482
- https://github.com/posit-dev/images-shared/issues/217


bschwedler:
## Scan-identifying metadata sent to `wizcli`

These are the fields from `WizCLICommand` (`posit-bakery/posit_bakery/plugins/builtin/wizcli/command.py`) that identify a scan in the Wiz UI: the scan **name**, the **context ID** it's grouped under, and the **tags** attached to it.

### Tags (`--tags`, repeated)

Auto-generated from `ImageTarget` via `scan_tags`, then any user-supplied `tool_options.tags` are appended after.

| Tag key | Source | Always present? | Example |
| --- | --- | --- | --- |
| `product` | `image_target.image_name` | always | `product=connect` |
| `version` | `image_target.image_version.name` | always | `version=2025.09.0` |
| `channel` | `image_target.release_channel.value` | always | `channel=release` |
| `platform` | `SETTINGS.architecture` | always | `platform=amd64` |
| `base-os` | tag template `OS` value | only if the image has an OS axis | `base-os=ubuntu-22.04` |
| `variant` | tag template `Variant` value | only if the image has a Variant axis | `variant=standard` |
| *(user-defined)* | `tool_options.tags` (bakery.yaml) | optional, appended last | `team=platform` |

`os` is a reserved tag key in Wiz, so the OS axis is emitted as `base-os` instead.

### Context ID (`--scan-context-id`)

Groups all platform scans (amd64 + arm64) of the same build together and lets Wiz update a scan in place instead of creating a new one every run. Computed by `default_scan_context_id`, unless overridden by an explicit `--scan-context-id`:

- **Non-release channels** (`daily`, `preview`): `{image_name}-{channel}` — one context per channel, always reflecting the latest build.
- **Release channel**: `{image_name}-{version with patch/build-number stripped to monthly granularity}` — patch bumps and build-number increments update the same context instead of creating a new one each time.

| Image | Channel | Version example | Context ID |
| --- | --- | --- | --- |
| `connect` | daily | — | `connect-daily` |
| `connect` | release | `2025.09.0` | `connect-2025-09` |
| `connect` | release (Positron build-number suffix) | `2026.08.1-2` | `connect-2026-08` |
| `connect-content` | daily | — | `connect-content-daily` |
| `connect-content` | release | `r4.5-python3.14` | `connect-content-r4-5-python3-14` |
| `package-manager` | preview | — | `package-manager-preview` |
| `package-manager` | release | `2025.09.0` | `package-manager-2025-09` |
| `workbench` | preview | — | `workbench-preview` |
| `workbench` | release | `2025.09.0` | `workbench-2025-09` |
| `workbench-session` | release | `r4.5-python3.14` | `workbench-session-r4-5-python3-14` |

### Name (`--name`)

Not part of the grouping key (that's the context ID), but shown here since it's the other identifying label in the Wiz UI. Format: `{image_name}:{Version}-{OS}-{Variant}-{platform}`, matching bakery's per-platform cache tag format — so it maps directly to a real build artifact.

- `connect:2025.09.0-ubuntu-22.04-standard-amd64`
- `workbench:daily-ubuntu-22.04-standard-arm64`

bschwedler:
@rstu-js Could you please provide feedback on the name, context id, and tags?

bschwedler:
## `--scan-context-id` granularity

I need input from someone with more Wiz knowledge. The current scheme groups scans
that are not comparable.

### What we send now

One context ID covers every OS, variant, and platform of a release. For the `connect`
image in `images-connect`, production settings give 100 scans across 14 context IDs:

| context-id | scans that share it |
|---|---|
| `connect-2025-05` … `connect-2025-11` | 4 each |
| `connect-2025-12` … `connect-2026-04` | 8 each |
| `connect-2026-05` … `connect-2026-07` | 12 each |

All 12 scans of 2026.07.0 send `--scan-context-id connect-2026-07`. They differ by
OS (3), variant (2), and platform (2).

### Why this can be a problem

Wiz documents the flag as a baseline selector. The baseline is the last scan on the
same repository, branch, and context ID. Two results follow.

1. The baseline is nondeterministic. Matrix jobs are `(image, version, platform)`, so
   amd64 and arm64 run in parallel and write the same context. Job order decides which
   scan becomes the baseline.
2. Cross-OS baselines can report false "resolved" findings. A CVE in 22.04 but not in
   26.04 looks fixed. Nothing was fixed. I did not test this against a tenant.

Weekly rebuilds are the case that matters. That cron exists to pick up upstream
base-image changes.

### Proposed change

One context ID per artifact. Keep the month-strip, so patch bumps continue a baseline:

```
--name              connect:2026.07.0-ubuntu-22.04-std-amd64
--scan-context-id   connect-2026-07-ubuntu-22-04-std-amd64
```

Tags keep the grouping. We already send `product`, `version`, `channel`, `base-os`,
`variant`, and `platform`.

### Questions

1. Does Wiz deduplicate findings inside a scan context, or is the context only a
   baseline selector? This is the best argument to keep the collapse. `std`/`min` and
   `amd64`/`arm64` overlap a lot.
2. Is there a limit on the length or character set of `--scan-context-id`? The new IDs
   reach about 55 characters.
3. Can a "resolved since baseline" signal exist that a cross-artifact baseline corrupts?
4. Is coarse grouping deliberate practice?

If the answer to 1 is yes, I will revisit the variant and platform axes. OS stays
split, because finding overlap is lowest there.

bschwedler:
At this point we have decided to keep `context-id` unset until we have the need for it in `wizcli diff` mode: https://docs.wiz.io/docs/diff-based-mode-wiz-cli

rstu-js (approved):
Ben & I discussed removal of --scan-context-id since that is only supposed to be used in diff mode. Other than that, this plan & implementation looks good to move forward. Security will be doing a full threat model on the complete epic & implementation to address any best practice findings & security risks.

PR #743, opened 2026-08-11
Restore pyflakes in the ruff gate and correct the Python floor

### LINKED ISSUE #742: Ruff configuration disables pyflakes and hides real defects
`posit-bakery/pyproject.toml` sets `[tool.ruff.lint] select = ["T201"]`. Ruff's `select` replaces the default rule set. It does not add to it. Only `T201` is active, so every other rule is off. Commit `a13d14ff` added this setting on 2024-11-22.

This is not a theoretical problem. 80 findings appear when pyflakes is active. Four of them are defects:

- `requires-python = ">=3.10"` is wrong. `posit_bakery/error.py` defines three classes that subclass the builtin `ExceptionGroup`. That class exists only from Python 3.11 (PEP 654). The metadata permits installation on Python 3.10. The CLI then stops with `NameError`, because `cli/build.py` imports `error.py`. The trove classifiers also advertise Python 3.10.
- Four tests never ran. A second definition with the same name shadowed each one. One is a parametrized `test_build_args` with about sixteen cases. A copy of `cli_cwd_context` also shadowed the `cli_tmpcontext` fixture.
- A call to an undefined function passed the lint step. `_emit_summary()` did not exist. Rule `F821` reports it.
- Nine bare `except:` clauses in CLI handlers catch `BaseException`. As a result, a handler catches Ctrl-C and shows it as a command failure.

mypy finds some of these problems. mypy does not run in CI.

bschwedler:
Ruff's `select` replaces the default rule set. It does not add to it. A configuration with only `T201` therefore turns off every other rule. This is why a call to an `_emit_summary()` that did not exist passed the lint step and reached CI.

This change writes the rule set out in full. It does not use `extend-select`. Ruff's default enables about four hundred rules and gives 629 findings that are unrelated to this change. An explicit list also means that a ruff upgrade cannot change what CI enforces without a commit.

The 80 findings included two defects. The declared Python floor was wrong, because `error.py` cannot import on Python 3.10. Four tests never ran, because duplicate definitions shadowed them. Each shadowed test is handled on its own terms, not by deletion of the duplicate. In one case the shadowed test was the stricter test of the pair.

One finding stays open. `wizcli/suite.py` assigned a `parse_err` value that nothing reads. The suite logs a malformed report but does not surface it when wizcli exits with code 0. This change removes the dead variable only. A change to when scans fail does not belong in a lint change.

Closes https://github.com/posit-dev/images-shared/issues/742

### DIFF
--- a/posit-bakery/posit_bakery/cli/ci.py
+++ b/posit-bakery/posit_bakery/cli/ci.py
@@ -370,7 +370,7 @@ def matrix(
 
     except typer.Exit:
         raise
-    except:
+    except Exception:
         log.exception("Failed to load bakery config")
         raise typer.Exit(code=1)
 
--- a/posit-bakery/posit_bakery/cli/create.py
+++ b/posit-bakery/posit_bakery/cli/create.py
@@ -146,7 +146,7 @@ def image(
             description=description,
             documentation_url=documentation_url,
         )
-    except:
+    except Exception:
         log.exception("Error creating image")
         stderr_console.print(f"❌ Failed to create image '{image_name}'", style="error")
         raise typer.Exit(code=1)
@@ -259,7 +259,7 @@ def version(
             latest=mark_latest,
             force=force,
         )
-    except:
+    except Exception:
         log.exception("Error creating version")
         stderr_console.print(f"❌ Failed to create version '{image_name}/{image_version}'", style="error")
         raise typer.Exit(code=1)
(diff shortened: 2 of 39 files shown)


ianpittwood (commented):
Are the key changes in posit-bakery/pyproject.toml?

bschwedler:
> Are the key changes in posit-bakery/pyproject.toml?

Yes, it is mostly a ruff config change.

We had to bump python to 3.11 because we are using `ExceptionGroup`.

ianpittwood:
I think this will need to be force merged as it will fail the linting checks in the merge queue.

